Issue #6, opened 2023-05-02
Setup Database for Enrolment Portal


b-f-chan:
Setup the Postgres SQL database and the Prisma ORM for the enrolment portal

anncatton:
Merged into develop. Since this is just docs/scaffolding, no need for usual testing pipeline

PR #18, opened 2023-06-12
Add dbname, prisma generate and migrate scripts

anncatton:
Additional work for #6 - Adds some configuration scripts for prisma migrations and updates README
- adds a Makefile, including command to run docker + migrations + dev servers
- adds `POSTGRES_DB` name to docker-compose psql config
- adds prettier dep & config
- updates README with local dev instructions
- updates `.env.schema` for api, creates one for ui
- adds health endpoint


apps/api/src/index.ts
```diff
@@ -1,13 +1,19 @@
 import express from 'express';
 import { PrismaClient } from '@prisma/client';
+import * as dotenv from 'dotenv';
 
 const prisma = new PrismaClient();
 const app = express();
+dotenv.config();
```
joneubank, line 7: i would probably move this to the top of a config file that handles reading in all the env stuff.

joneubank (approved):

PR #21, opened 2023-06-27
🗃️ 11 - add participant, question and response schemas

anncatton:
Renames `User` model to `Participant`. Deletes previous init migration that created the `User` table
Adds `ConsentQuestion` and `ParticipantResponse` models + migrations.
- the `on delete` and `on update` relationship for a participant response with participants and consent questions is currently set to *not* perform any cascading changes. This may need to be revisited depending on requirements.


apps/api/prisma/migrations/20230525183353_init/migration.sql
mistryrn, line 1: #rip

apps/api/prisma/schema.prisma
```diff
@@ -7,8 +7,36 @@ generator client {
   provider = "prisma-client-js"
 }
 
-model User {
+model Participant {
```
mistryrn, line 10: 🥲

mistryrn (approved):
Good stuff! 👍

apps/api/prisma/migrations/20230525183353_init/migration.sql
anncatton: lol...we hardly knew ye

PR #22, opened 2023-06-28
📝 add readme for prisma migrations

anncatton:
Starting a (hopefully helpful) doc to track issues encountered with prisma migrations!


mistryrn (approved):
This is great! Really easy to read, example makes sense and is easy to follow. Thanks for putting this together :)

PR #23, opened 2023-06-30
➕🔧🎨 12 - Routers setup, init swagger

anncatton:


apps/api/src/config.ts
mistryrn, line 1: 👏 So clean!

apps/api/src/index.ts
mistryrn, line 1: 🤩  It's so organized!

apps/api/src/prismaClient.ts
mistryrn, line 1: Thanks for pulling this out into a separate file!

apps/api/src/routers/consentQuestions.ts
```diff
@@ -0,0 +1,102 @@
...
+const router = Router();
+
+/**
+ * @openapi
+ * tags:
+ *   - name: Consent Questions
+ *     description: Consent Questions management
+ */
+
+/**
```
mistryrn, line 36: These docs are so handy!

apps/api/src/service/search.ts
```diff
@@ -0,0 +1,57 @@
+import { Request } from 'express';
+import prisma, {
+  ParticipantResponse,
+  Participant,
+  ConsentQuestion,
+} from '../prismaClient';
+
+// TODO: implement with some good defaults. do we need pagination etc?
```
mistryrn, line 8: Good question re: pagination 🤔 I don't have the answer for now, but I'm sure as we build this out it'll become clear the defaults that we need.

apps/api/src/service/search.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const getConsentQuestions = async (
+  parameters: any = {}
+): Promise<ConsentQuestion[]> => {
+  const result = await prisma.consentQuestion.findMany({ where: parameters });
+  return result;
+};
+
+export const getLatestParticipantResponseByParticipantIdAndQuestionId = async (
```
mistryrn, line 43: 👏

mistryrn (approved):
This all looks great! I'll respond to the individual TODO items from the PR description in a follow-up comment, but I approve of all of this stuff. Nice and organized, and the docs look dope!

mistryrn:
Just going through the TODOs/separate ticket things from the PR description:

>    Using Prisma model types for type-checking in other files (exported from prisma client), these should be handled in the common types with zod (?)

Yes, we'll want to get the common zod stuff set up/updated as we build this out. I think Zod has some kind of Type exporter which looked really handy when I was setting up the monorepo. I can pull up my notes on this if you want to reference them when working on this.

Agreed, let's make a separate ticket for this. ✅ 

>    implement proper health check in health router

Agreed ✅ 

>    error handler (incl. handling prisma error types)

Agreed ✅ 

>    auth filter

Agreed ✅ Though I'm not sure how/when we want to tackle this, seeing as we don't have auth setup just yet. What do you think?

>    reusable swagger schema components

I did not know this was a thing, but agreed! ✅ 

>    proper logging

Agreed ✅ Any preference on logging libraries?

>    fetching application version from package.json

Agreed ✅ 

>    default parameters for db queries (helper func)(?)

Agreed ✅  This is a good idea. What do you have in mind here?

>    graceful server shutdown

Agreed ✅

apps/api/src/prismaClient.ts
anncatton: i was checking out their docs here https://www.prisma.io/docs/guides/performance-and-optimization/connection-management , looks like what they recommended for our case - i guess if we need to get fancy database-wise we can revisit!

PR #24, opened 2023-07-10
14 init i18n

anncatton:


apps/ui/src/app/[lang]/page.tsx
```diff
@@ -0,0 +1,39 @@
+import Link from 'next/link';
+import { User } from 'common';
+
+import { getDictionary, useTranslation } from '../i18n';
+import Header from '../components/Header';
+
+const user: User = {
+  id: '1',
+  name: 'Homer Simpson',
+  email: 'homersimpson@example.com',
```
Comment (justincorrigible) [inline apps/ui/src/app/[lang]/page.tsx:10]: classic

apps/ui/src/app/[lang]/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export default async function Home({
+  params: { lang },
+}: {
+  params: { lang: string };
+}) {
+  const { t } = await useTranslation(lang);
```
Comment (justincorrigible) [inline apps/ui/src/app/[lang]/page.tsx:18]: Inclined to suggest that `t` be renamed `t8` if only just to make it easier to find

apps/ui/src/app/components/LanguageToggle/LanguageList.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { getDictionary, useTranslation } from '../../i18n';
+
+export default async function ({ lang }: { lang: string }) {
+  const { t } = await useTranslation(lang, 'header');
+  const dict = await getDictionary(lang, 'header');
+  return (
+    <ul>
+      {supportedLanguages
+        .filter((lng) => lang !== lng)
+        .map((l) => {
```
justincorrigible, line 13: This component confuses me a little. Please bear with me as I try to reason it "out loud" in this comment: 
in the context of a "Language List" (e.g. as intended in `<ul/>`), you would expect multiple elements to be shown. right? Multiple "supported languages", in our case.
Then, there's this filter. Do we anticipate the `lang` string in the props will match more than one supported language? If so, that would imply duplicates in the `supportedLanguages` array, which doesn't seem likely.
Thinking there will only be one possible match, `supportedLanguages.find` may be more declarative than `.filter().map()`.  And then, just for safety, I'd implement a fallback for no matches (English?), to be shown instead of a "list" with either one or no elements. 

But if it is either 1 or 0 elements... is it really ever a list?

🤔 Somehow I feel I'm just missing something here.

apps/ui/src/app/[lang]/page.tsx
Comment (anncatton) [reply apps/ui/src/app/[lang]/page.tsx:10]: that's all @mistryrn ...but of course i couldn't remove it lol

apps/ui/src/app/components/LanguageToggle/LanguageList.tsx
anncatton: tbh the list setup is just from the blog/docs i was reading from. def not in its final form here, you're right that in our case it will be an either/or scenario
For the fallback language, that *should* basically be handled by the middleware and settings files, at the routing level. One thing that came up with that is that, with the current setup, if you were to put in an unsupported locale like `de`, the middleware would treat that as `<root>/<pathname="de">` and so return a url of `/en/de`, which would give a 404. This behaviour can be modified with the `matcher` in the [middleware file](https://github.com/OHCRN/consent-platform/pull/24/files#diff-36b521ade291c54cfa72032163cd49784107a33cab0461c330facd431be8689aR24)...but atm i'm not exactly sure what behaviour i'd want there. I'm open to ideas on that one!

apps/ui/src/app/[lang]/page.tsx
Comment (justincorrigible) [reply apps/ui/src/app/[lang]/page.tsx:18]: agreed 100%. thank you

apps/ui/src/app/components/LanguageToggle/LanguageList.tsx
justincorrigible: oh fun misbehaviour. almost makes me suggest we switch away from `ohcrn.com/en` to a `en.ohcrn.com` but I'm not 100% sure right now what complexity that would bring, or whether that has trade-offs. 

Already doing lots of reading today on this whole "new" routing in NextJS, so that I can offer actual usable insight instead of just ideas. TBD.

apps/ui/src/app/[lang]/page.tsx
Comment (justincorrigible) [reply apps/ui/src/app/[lang]/page.tsx:10]: don't you dare!

apps/ui/src/app/components/LanguageToggle/LanguageList.tsx
anncatton: The preference from OHCRN is for the `/[lang]` route syntax, but i think because we have control of the middleware we can decide what kind of redirect behaviour we want!

PR #30, opened 2023-07-19
🎨 🌐 move i18n code outside app dir, add types

anncatton:
Reorganizes i18n dir. Will keep i18n support in the consent platform UI package as it may or may not be reusable in future
- moves this code outside of the `app` folder
- adds typing


apps/ui/src/i18n/hooks/getTranslation.ts
```diff
@@ -17,40 +17,24 @@
...
  */
 
-import {
-  ValidLanguage,
-  ValidNamespace,
-  defaultLanguage,
-  defaultNS,
-} from './settings';
+import { GetDictionary, GetTranslation } from '../types';
+import { defaultLanguage, defaultNamespace } from '../settings';
```
joneubank, line 21: If we're doing absolute paths let's be consistent and not have relative paths...

joneubank (approved):
Is good. Just be consistent ;)

PR #32, opened 2023-07-21
💄🎨 - 15 basic layout components

anncatton:
Adds basic page layout, header and footer + basic theming.
- adds sass dep
- adds SVG files for header and footer icons
- adds global header, root element styles
- updates tailwind config with style guide colours and fonts
- removes tailwind boilerplate styling
- adds `footer` translation namespace
- adds util function for merging css classnames
- fixes some remaining relative imports
- adds favicon to middleware matcher to prevent appending `lang` to the image url

Notes:
- icon links use `#` url for now, will need to add valid links in dictionaries
- API and Registry versions are hardcoded for dictionary params, will need update to retrieve correct versions
- OICR, OHCRN and ON govt svgs are currently english versions only, will need to add French versions


apps/ui/src/app/[lang]/page.tsx
```diff
@@ -17,19 +17,13 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import Header from '../components/Header';
 import { ValidLanguage } from '@/i18n';
-import HomeComponent from '../components/Home';
+import HomeComponent from '@/components/Home';
```
Comment (justincorrigible) [inline apps/ui/src/app/[lang]/page.tsx:21]: 🎉 yuusssss!!!

apps/ui/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+        </Link>
+        <Link href="#" className={styles.icon}>
+          <Image src={GithubLogo} alt={translate('github-alt')} />
+        </Link>
+      </div>
+      <div className={styles.copyright}>
+        <span>{translate('copyright')} </span>
+        <span>
+          {/* TODO: fix hardcoded version */}
+          {translate('ohcrn-registry', { registryVersion: '0.1.0' })} -{' '}
```
justincorrigible, line 81: fancy!

apps/ui/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+          src={TwitterLogo}
+          alt={translate('twitter-logo-alt')}
+          className={styles.twitter}
+        />
+      </Link>
+    </div>
+  );
+};
+
+const Footer = async ({ lang }: { lang: ValidLanguage }) => {
```
justincorrigible, line 144: Inclined to suggest the other components/functions should be isolated into their own files (e.g. `Footer/Links.tsx`, but I'm not sure whether that makes things easier or is just additional boilerplate.

Thoughts?

apps/ui/src/components/PageLayout/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  lang: ValidLanguage;
+}) => {
+  return (
+    <>
+      <Container>
+        <Header lang={lang} />
+        <main className={styles.main}>{children}</main>
+        <Footer lang={lang} />
+      </Container>
+    </>
```
justincorrigible, line 40: since all those components are withing the one `Container`, we don't need a `Fragment` here.
(this suggestion is for illustration only, as it doesn't re-indent things)
```suggestion
      <Container>
        <Header lang={lang} />
        <main className={styles.main}>{children}</main>
        <Footer lang={lang} />
      </Container>
```

apps/ui/src/global/utils/mergeClassnames.ts
```diff
@@ -0,0 +1,2 @@
+export default (...args: (string | undefined)[]): string =>
+  args.filter(Boolean).join(' ');
```
justincorrigible, line 2: you may want to check this one out instead ;) https://www.npmjs.com/package/classnames

apps/ui/src/components/Footer/index.tsx
anncatton: That's a good idea, this file is getting rather busy. i wanted to get some eyes on this pr before i proceeded further though, it was getting out of hand! I'll extract them to their own files before merging

apps/ui/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+import OICRLogo from '@/public/oicr.svg';
+import InstagramLogo from '@/public/instagram.svg';
+import TwitterLogo from '@/public/twitter.svg';
+import ONGovtLogo from '@/public/on_govt.svg';
+import GithubLogo from '@/public/github.svg';
+import OvertureLogo from '@/public/overture.svg';
+
+import mergeClassnames from '@/global/utils/mergeClassnames';
+
+const footerLinks: { translation: string; url: string }[] = [
```
mistryrn, line 36: I think you and Justin have already touched on this below, but since this file is getting a bit busy it's probably a good idea to move some of this stuff into their own files. Maybe a globals/constants for URLs/routes/links could work for these footer links?

apps/ui/src/components/PageLayout/index.tsx
mistryrn, line 1: It's so clean 🥲

apps/ui/tailwind.config.js
```diff
@@ -1,17 +1,32 @@
...
     './src/components/**/*.{js,ts,jsx,tsx,mdx}',
     './src/app/**/*.{js,ts,jsx,tsx,mdx}',
   ],
   theme: {
+    colors,
+    fontFamily: {
+      sans: ['var(--font-montserrat)']
+    },
+    fontSize: {
+      xs: ['13px', {lineHeight: '15.85px', fontWeight: '400'}],
```
mistryrn, line 17: How do we feel about switching the font-size base units to `rems` instead of `px` for accessibility? So that when someone zooms in their browser, the text scales relatively?

Not sure if Figma has unit conversion/toggles though. May require a bit of digging to find out what the base is (if it's not just the default) and then some math to go from like 24px to 1.5rem (if the base is 16px for example)

mistryrn (approved):
Looks great! Just had a couple comments regarding breaking out some things from the Footer, and switching font units from `px` to `rem` :)

apps/ui/tailwind.config.js
anncatton: great call! i can check with @patrickdsantos about this.

justincorrigible: At risk of stating the obvious, the size for all of these can be respected as they are.
We can just calculate them as `"current font size in px" /  "the html/body font size"`
e.g. if the base font is 16px, then 13px would be 13px/16px -> 0.8125rem -> 0.81 or 0.82 rem

justincorrigible (approved):

apps/ui/tailwind.config.js
anncatton: That sounds reasonable to me (was not however quite obvious to me 😛 ). Looks like tailwind has some defaults we might be able use: https://tailwindcss.com/docs/font-size

justincorrigible: While I'm not 100% comfortable relying on third party libraries to define my styles (PTSD from years of Twitter Bootstrap hell), I won't be the one working on this app, so I'll leave the choice and details up to you.

I trust your judgment 🙏.

anncatton: Our font sizes differ slightly from tw, so i'll keep the custom config i've added in this pr for those specific sizes and add the rem sizing! So hopefully that eases the ptsd somewhat 😛

justincorrigible: The voices are quieter now. lol
thank you! 🙏

anncatton:
Left a `TODO` in the tailwind config but adding a note here as well: will make the fontsize changes from `px` to `rem` in a separate PR

Issue #13, opened 2023-05-24
Setup LINTER for Custom NextJS App for Enrolment Portal


b-f-chan:
Customize the default NextJS app for the OHCRN Enrolment Portal

- [ ] Add linter, prettier, other tools required
- [ ] Setup directory structure for components and utils as required

b-f-chan:
Adding @anncatton jointly to this ticket to investigate while @mistryrn is away

b-f-chan:
@anncatton will familiarize herself with Next13 --> May need to touch base with Rakesh.  Just create container for webapp and set to NOT default

anncatton:
added `prettier` in #18

b-f-chan:
@mistryrn - To take time this sprint to take a stab at the linter

b-f-chan:
@mistryrn - Please aim to close this one off before starting Data Model implementation

b-f-chan:
Merged, closing

PR #39, opened 2023-07-29
🔧 Configure ESLint, Update Prettier Config, Add Workspace Config (#13)

mistryrn:
* Add custom shared ESLint config in `packages/config/eslint-config-ohcrn`
* Update Prettier config
* Remove redundant `pnpm-lock.yaml`s from `api` and `ui` projects (root `pnpm-lock.yaml` is the only one needed)
* Update `pnpm-workspace.yaml` to check for workspaces in subdirectories of the `packages` directory
* Update README to add VS Code setup instructions
* Update `.gitignore` to allow certain VS Code workspace config files
* Add VS Code recommended extensions in `.vscode/extensions.json` and recommended settings in `.vscode/settings.schema.json`* Update `.gitignore` to allow certain VS Code workspace config files


README.md
```diff
@@ -62,6 +62,23 @@ To run the setup locally, ensure you have provided the **required** environment
 - Install pnpm: `brew install pnpm`
 - Install dependencies: `pnpm install`
 
+### VS Code Configuration
+
+[Visual Studio Code](https://code.visualstudio.com/) is the recommended code editor for this project. The monorepo contains a [`.vscode` directory](./.vscode/) with a [`settings.schema.json` file](./.vscode/settings.schema.json).
+
+Please **create a copy of the `settings.schema.json` file called `settings.json`**. This will configure VS Code with our recommended settings and enable ESLint fix-on-save, but leaves you free to configure your workspace settings to your preference without committing any changes to the repo.
```
anncatton, line 69: Niiice

packages/config/eslint-config-ohcrn/index.js
```diff
@@ -0,0 +1,63 @@
...
+			{
+				groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
+				'newlines-between': 'always',
+				warnOnUnassignedImports: true,
+			},
+		],
+		'import/newline-after-import': 'warn',
+		'prettier/prettier': [
+			'warn',
+			{
```
anncatton, line 41: out of curiosity is it possible to reuse one prettier config here and for the prettier rc (maybe both as json) so they're always in sync? i have not tried this, just wondering 😛

packages/config/eslint-config-ohcrn/index.js
```diff
@@ -0,0 +1,63 @@
+module.exports = {
```
anncatton, line 1: i noticed in the linting pr that an unused import gets commmented out - what do you think of setting it up to delete unused imports instead?
https://github.com/OHCRN/consent-platform/pull/40/files#diff-c965c03a8feac148581c1f7417dad111a48c19d33b276f0e96292239e7f850a7R25

README.md
```diff
@@ -62,6 +62,23 @@ To run the setup locally, ensure you have provided the **required** environment
...
+In addition to downloading VS Code, please ensure that you have the following VS Code Extensions:
+
+- [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
+- [Prettier - Code formatter](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)
+
+> **Note:** These extensions are also listed in the `recommendations` of the [`extensions.json` file](./.vscode/extensions.json)
+
+### Special Considerations Regarding ESLint and the Root `package.json` File
+
+Please note that we are currently specifying versions of `@typescript-eslint/parser` and `eslint-plugin-import` in the `overrides` section of the root [`package.json` file](./package.json). This was done to avoid a conflict between the version used by the `eslint-config-next` plugin and our custom ESLint configuration.
```
anncatton, line 80: 👍

anncatton (approved):
This looks great! Appreciate those notes in the README, future me thanks you 🙏 

Left a couple comments, feel free to ignore as it's just me thinking out loud 🙂

.gitignore
```diff
@@ -58,7 +58,9 @@ report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
 
 # ignore vscode workspace settings
 *.code-workspace
-.vscode
+.vscode/*
+!.vscode/extensions.json
```
justincorrigible, line 62: 👍 to extension suggestions

.vscode/settings.schema.json
```diff
@@ -0,0 +1,17 @@
...
+  "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": true
+  },
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    "typescript",
+    "typescriptreact"
+  ],
+  "editor.tabSize": 2,
```
justincorrigible, line 15: I may suggest we leave this out, as I'm inclined to think it's perhaps just "for spaces" (not tabs), and I personally prefer seeing my tabs as 4 spaces long anyway, which should be optional

README.md
```diff
@@ -62,6 +62,23 @@ To run the setup locally, ensure you have provided the **required** environment
...
+### VS Code Configuration
+
+[Visual Studio Code](https://code.visualstudio.com/) is the recommended code editor for this project. The monorepo contains a [`.vscode` directory](./.vscode/) with a [`settings.schema.json` file](./.vscode/settings.schema.json).
+
+Please **create a copy of the `settings.schema.json` file called `settings.json`**. This will configure VS Code with our recommended settings and enable ESLint fix-on-save, but leaves you free to configure your workspace settings to your preference without committing any changes to the repo.
+
+In addition to downloading VS Code, please ensure that you have the following VS Code Extensions:
+
+- [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
+- [Prettier - Code formatter](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)
```
justincorrigible, line 74: if ESLint is running prettier as a plugin, is the VSCode prettier extension still needed?

package.json
```diff
@@ -29,5 +29,11 @@
     "sinon": "^15.2.0",
     "ts-node": "^10.9.1",
     "typescript": "^5.1.6"
+  },
+  "pnpm": {
+    "overrides": {
```
justincorrigible, line 34: le fancy 💅

packages/config/eslint-config-ohcrn/index.js
justincorrigible: That sounds reasonable. is it doable?

packages/config/eslint-config-ohcrn/index.js
mistryrn: Oh good call! I was supposed to look into this 😅

packages/config/eslint-config-ohcrn/index.js
mistryrn: So this commenting wasn't actually done by the linter, but by me to avoid removing some of your code which may have been important or part of something you were still working on since this was just one big sweeping `--fix-all`

It currently will just throw an error and give a big red underline for unused imports or variables. Do we still want to set it up to delete them? Keep in mind that this would run on save as well

.vscode/settings.schema.json
mistryrn: That's a good call, we can leave the tab size up to the individual developer!

README.md
mistryrn: This is a good question. I'll test with and without Prettier disabled and see if there's any difference/duplication!

packages/config/eslint-config-ohcrn/index.js
mistryrn: Update on this: unfortunately after some digging and trying different combinations of JSON, JS (ESM), JS (CommonJS) as the [config file](https://prettier.io/docs/en/configuration.html) for prettier and using import or require in the eslint config, it either breaks the eslint config file or fails to load the external prettier config :(

I'm gonna park my investigation into that for now, but if anyone else wants to try it later I'd be happy to discuss the combinations I tried!

justincorrigible: thanks for looking into this!
we can leave it as a ToDo for sure

Issue #34, opened 2023-07-25
Add Translation Middleware for all Routes in Web App


b-f-chan:
Add the French translation middleware setup for all routes in the Web App

Content will be done in separate ticket once translation company provides it

b-f-chan:
@anncatton - Additional scope as agreed for this ticket, simplify the routes:

We will create a separate ticket to revise and simplify the URLs

We will keep the subdomains
We will go with paths not parameters
The URLS will be renamed as follows:
Homepage: https://registry.ohcrn.ca --> STAYS SAME
Participant Registration: https://registry.ohcrn.ca/en/participant/registration --> CHANGE TO /register
Participant Keycloak Create Account: https://keycloak.ohcrn.ca/auth/realms/virusseq/login-actions/registration --> STAYS SAME
Clinician Patient Registration : https://registry.ohcrn.ca/en/clinician/registration --> CHANGE TO /invite
Participant Dashboard: https://registry.ohcrn.ca/en/participant/dashboard --> CHANGE TO /dashboard
Consent Wizard: https://registry.ohcrn.ca/en/participant/consent-forms --> CHANGE to /consent

b-f-chan:
Just needs minor/quick re-review of PR

b-f-chan:
Merged, closing

PR #46, opened 2023-08-03
✨🔥 33 - add clinician, participant routes.

anncatton:
Add routing for clinician, participant routes
- index pages are just boilerplate
- adds links from home page for demo purposes
- paths as constants will be implemented in https://github.com/OHCRN/platform/issues/34


mistryrn (approved):
Looks good! Now we have real routes, yay! 🎉

PR #47, opened 2023-08-03
💄 16 - add mobile lang toggle

anncatton:
Styling/display for language toggle on mobile screen
- moves toggle component to header dir

### DIFF
--- a/apps/consent-ui/src/components/Header/LanguageToggle.tsx
+++ b/apps/consent-ui/src/components/Header/LanguageToggle.tsx
@@ -19,39 +19,29 @@
 
 'use client';
 import { usePathname } from 'next/navigation';
+import { ReactNode } from 'react';
 
-import { defaultLanguage } from '@/i18n/settings';
 import { ValidLanguage } from '@/i18n';
-import { getUnselectedLang } from '@/components/Header';
 import LinkButton from '@/components/Button/LinkButton';
 
-import styles from './Header.module.scss';
-
-const replaceLocaleInUrl = (path: string, currentLang: ValidLanguage = 'en'): string => {
-	const unselectedLang = getUnselectedLang(currentLang);
-	const splitPath = path.split('/');
-	if (!splitPath.length) {
-		return defaultLanguage;
-	}
-	const originalLangRemoved = splitPath.slice(2).filter((item) => item.length);
-	const joinedPath = [unselectedLang.toString()].concat(originalLangRemoved).join('/');
-	return joinedPath;
+const getPathWithNewLocale = (locale: ValidLanguage, pathname?: string) => {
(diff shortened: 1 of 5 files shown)


apps/consent-ui/src/components/Header/LanguageToggle.tsx
```diff
@@ -25,30 +25,33 @@ import { ValidLanguage } from '@/i18n';
 import { getUnselectedLang } from '@/components/Header';
 import LinkButton from '@/components/Button/LinkButton';
 
-export const replaceLocaleInUrl = (path: string, currentLang: string = 'en'): string => {
-	const unselectedLang = getUnselectedLang(currentLang as ValidLanguage);
+import styles from './Header.module.scss';
+
+const replaceLocaleInUrl = (path: string, currentLang: ValidLanguage = 'en'): string => {
```
justincorrigible, line 30: TBH, I'm not sure I love this url replacement logic. 🤔
Thinking perhaps something like `url.replace(regex)` could do a cleaner job?

mistryrn: I think because originally we were designing this to support the possibility of 3+ languages, but have since pivoted to focusing on the 2 languages we know we'll be supporting for sure, this logic feels a bit complicated now.

Looking at some of the props the LanguageToggle component and its parent [Header](https://github.com/OHCRN/platform/blob/b4cbe7278d8ed58432d9781357c0388be58033a8/apps/consent-ui/src/components/Header/index.tsx) component have/are expecting (LanguageToggle: `currentLang`, `langToSelect`; Header: `lang`), we might be doing some duplicated work here to determine the language to replace.

Maybe something similar to the basic Next.js example here could work? https://github.com/vercel/next.js/blob/d7405b0c288a0b8974b20ff7de8c291d02f22710/examples/app-dir-i18n-routing/app/%5Blang%5D/components/locale-switcher.tsx#L9

Another possibility that I only just discovered (and may not be applicable for our needs, I haven't tried this yet and don't know how well it handles url params) is this `locale` prop that `next/link` and `next/router` support: https://nextjs.org/docs/pages/building-your-application/routing/internationalization#transition-between-locales

anncatton: @justincorrigible i used the example rakesh posted as for me that feels easier to read than involving a regex, but let me know what you think!

justincorrigible: 💯 great solution! Thanks both

justincorrigible (approved):

mistryrn (approved):
Looks good, thanks for updating that! 😄

PR #48, opened 2023-08-22
POC: environment variables

samrichca:
prototype for environment variables & data fetching


apps/consent-ui/src/app/[lang]/participant/consent-forms/ConsentButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+'use client';
+
+import { useAppConfigContext } from '@/components/AppConfig';
+import Button from '@/components/Button';
+
+const ConsentButton = ({ isComplete }: { isComplete: boolean }) => {
+	const appConfig = useAppConfigContext();
```
Comment (samrichca) [inline apps/consent-ui/src/app/[lang]/participant/consent-forms/ConsentButton.tsx:26]: getting an environment variable in a client component

.vscode/extensions.json
```diff
@@ -1,6 +1,7 @@
...
-  "recommendations": [
-    "dbaeumer.vscode-eslint",
-    "esbenp.prettier-vscode"
-  ]
-}
\ No newline at end of file
+	"recommendations": [
+		"dbaeumer.vscode-eslint",
+		"esbenp.prettier-vscode",
+		"bradlc.vscode-tailwindcss"
```
samrichca, line 5: this tailwind plugin is nice (rakesh approved)

apps/consent-ui/src/components/AppConfig/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+const defaultContextValues = {
+	BASE_URL: 'http://localhost:3000',
+	TEST_RUNTIME_VAR: 'testingtesting',
+	CONSENT_API_URL: 'http://localhost:8080',
+	FEATURE_FLAG: true,
+	NEXT_IS_BUILDING: false,
+};
+
+const AppConfigContext = createContext({} as AppConfig);
+const DefaultAppConfigContext = createContext(defaultContextValues as AppConfig);
```
ciaranschutte, line 35: no need to create 2 seperate contexts - just give the one context different values

ciaranschutte (changes requested):

apps/consent-ui/src/components/AppConfig/index.tsx
samrichca: i'm not sure what to do with line 48-52 in that case

ciaranschutte: Best to set the value of context that you created, instead of creating new contexts.
Think of a context as single instance of a state -  you can _set_ or _get_ anywhere in your React tree
ref: https://react.dev/learn/passing-data-deeply-with-context#

Your context in this case is something like `AppConfig`, holding app configuration info.
The intent is the default value of `AppConfig` should be `defaultContextValues` and then on line 58 we update `AppConfig` value with the result of our "get runtime var" code.

Currently it's creating two individual contexts.


An example from `rdpc-ui`
https://github.com/icgc-argo/rdpc-ui/blob/develop/src/app/hooks/AppProvider.tsx#L40

ciaranschutte (changes requested):

ciaranschutte:
verified runtime vars with a single `npm run build`
then running `npm run start` => TEST_RUNTIME_VAR is default value
then running `TEST_RUNTIME_VAR=working npm run start` => TEST_RUNTIME_VAR is "working"

ciaranschutte (approved):

ciaranschutte:
@samrichca couple of suggestions but runtime variables are working

Issue #49, opened 2023-08-29
Create DASes in the OHCRN Platform Repo


b-f-chan:
Create DASes in OHCRN/platform repo

Will involve porting work from DAS POC into the OHCRN/platform repo

- [x] PI DAS
- [x] PHI DAS
- [x] Keys DAS
- [x] Consent DAS

May want to have Private repo settings configured, though not a huge deal at this point in the implementation since the security aspects have yet to be implemented (CC: @justincorrigible )

Pretty low effort, mostly copy+paste from the POC

**EXIT CRITERIA:**

- [x] The /apps directory in the monorepo contains a DAS for each of the DASes listed above
- [x] The /apps directory in the monorepo contains a Data Mapper app
- [x] The /packages directory in the monorepo contains a shared Logger
- [x] The root `package.json` file is updated with scripts to run the new services

b-f-chan:
Logger and data mapping layers were also added to the ticket scope as agreed

b-f-chan:
May need to break this down into separate PRs for review, however work is complete

b-f-chan:
Reviewed, merged, closing

PR #64, opened 2023-09-08
✨ Add Data Mapper (#49)

mistryrn:
* Add initial version of Data Mapper service to `apps/` dir
* Update root `package.json` with `data-mapper-dev` command to run the Data Mapper


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "pnpm -r run build",
     "das-dev": "pnpm --filter '*-das' run dev",
+    "data-mapper-dev": "pnpm --filter data-mapper run dev",
```
anncatton, line 9: 👍

anncatton (approved):
I have nothing more insightful to add than "this is awesome" 😛 Thanks for all the thorough work @mistryrn !

apps/data-mapper/src/routers/participants.ts
```diff
@@ -0,0 +1,57 @@
...
+// get participant by id
+router.get('/:participantId', async (req, res) => {
+	logger.info('GET /participant/:participantId');
+	const { participantId } = req.params;
+	try {
+		const participant = await getParticipant(participantId);
+		res.status(200).send({ participant });
+	} catch (error) {
+		logger.error(error);
+		res.status(404).send({ error: 'Participant not found' });
```
justincorrigible, line 38: may I start a 204 vs 400 vs 404 debate? (jk)

apps/data-mapper/src/service/create.ts
```diff
@@ -0,0 +1,111 @@
...
+// KEYS-DAS
+// TODO: add Type instead of any
+const createParticipantOhipKey = async (participantId: string): Promise<any> => {
+	const { keysDasUrl } = getAppConfig();
+	// TODO: use urlJoin
+	// TODO: add error handling
+	const result = await fetch(`${keysDasUrl}/ohip-keys`, {
+		method: 'POST',
+		body: JSON.stringify({ participantId }),
+		headers: { 'Content-Type': 'application/json' },
```
justincorrigible, line 68: using axios, these headers can be abstracted to an axios instance. DRY

apps/data-mapper/src/service/search.ts
```diff
@@ -0,0 +1,61 @@
+// TODO: fix urlJoin error
+// import urlJoin from 'url-join';
+
+import { getAppConfig } from '../config';
+
+// combines data from respective DASes to return a single Participant object
+// TODO: add Type instead of any
+export const getParticipant = async (participantId: string): Promise<any> => {
```
justincorrigible, line 8: I'd advocate for moving this function's definition to the end of the file, so it doesn't invoke the other functions before they are defined (lines 9-12).

While hoisting may make this moot, code is more readable when variables are defined before they are used.

apps/data-mapper/src/service/create.ts
mistryrn: Great idea, thank you! I believe we are transitioning to using `axios` for HTTP requests going forward in this project (Alyssa and Sam are starting to use it in our `consent-ui` workspace), so I'll add some additional `TODO`s and create a ticket to track the work for switching from `fetch` to `axios` :)

apps/data-mapper/src/service/search.ts
mistryrn: Another good catch! Will update shortly 🙇

apps/data-mapper/src/service/create.ts
mistryrn: Ticket created to track fetch -> axios transition work here: https://github.com/OHCRN/platform/issues/72

PR #88, opened 2023-09-12
✨ Fetch consent-api version in footer

alyssadsouza:
- Added a `/status` endpoint for the consent-api status router to fetch the API version # from the package.json, which is then displayed in the consent-ui footer
- Installed Axios and introduced some new file structure in consent-ui for data fetching:
```
.
└── consent-ui/
    └── src/
        ├── app
        ├── config/
        │   ├── appConfig.ts
        │   └── index.ts
        ├── constants/
        │   ├── externalPaths.ts
        │   └── index.ts
        └── services/
            └── api/
                ├── axiosClient.ts
                ├── status.tsx
                └── index.tsx
```
  - `config` contains environment variable config stuff
  - `constants` is for string/int constants, etc.
  - `services/api` contains the base API instance and any calls to API endpoints using said instance


ciaranschutte:
one thing to be aware of with variables based on env vars eg. `apps/consent-ui/src/constants/externalPaths.ts`
this file will be statically built unless there's something preventing this

!remind me for more deetails

apps/consent-api/src/routers/health.ts
```diff
@@ -47,4 +50,23 @@ router.get('/', async (req, res) => {
 	res.json({ message: `API is running.` });
 });
 
+/**
+ * @openapi
+ * /health/version:
```
anncatton, line 55: Had a discussion with @mistryrn on this, we're thinking it would be good to rename this router to `status` instead of `health`, and include both the health check (can leave hardcoded for now like in the base endpoint, as we don't have a real health check yet) and the version in the same endpoint, like `{ status: 'API is healthy', version: packageVersion }`. Apologies, I think you asked this exact question while doing the work on this ticket! Then we only need the base endpoint from this file, and we'll leave the `TODO` for the healthcheck.

apps/consent-ui/package.json
```diff
@@ -13,6 +13,7 @@
     "@types/react": "18.2.14",
     "@types/react-dom": "18.2.6",
     "autoprefixer": "10.4.14",
+    "axios": "^1.5.0",
```
anncatton, line 16: 👍

apps/consent-ui/src/constants/externalPaths.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// API
+export const API_ROOTS = {
+	HEALTH: '/health',
+};
+
+export const API = {
+	BASE_URL: NEXT_PUBLIC_CONSENT_API_URL,
+	HEALTH: API_ROOTS.HEALTH,
+	VERSION: API_ROOTS.HEALTH + '/version',
```
anncatton, line 31: For url concatenation, we like to use [url-join](https://www.npmjs.com/package/url-join) - it prevents missing/extra `/`s... a lifesaver! 😂 
It's already in the package.json so you can just import as `import urlJoin from 'url-join`

packages/common/src/service/Health/Health.ts
```diff
@@ -0,0 +1,3 @@
+export type Version = {
+	version: string;
+}
```
anncatton, line 3: Seems like our linter/prettier is not adding newlines/formatting on save 🤔 Can you add a newline to the end of the file, and one of use can investigate why the formatter isn't working!

packages/common/src/service/Health/index.ts
```diff
@@ -0,0 +1 @@
+export * from './Health';
```
anncatton, line 1: same as above 🙂

apps/consent-api/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src",
+    "resolveJsonModule": true,
```
anncatton, line 5: i can see the nodemon complaints that you mentioned when i tried non-relative imports in the api...i'll poke around and see if i can figure out what's going on!

anncatton (commented):
Great work! Just leaving some general comments for now that aren't related to all the env var stuff. As you mentioned there'll be likely changes from that, esp on the data fetch stuff, so let's leave the PR open for the time being.
Actually, let's convert this to a draft PR for now until everything is sorted! 🙂

apps/consent-api/tsconfig.json
alyssadsouza: fixed in #89!

apps/consent-api/src/routers/status.ts
```diff
@@ -18,24 +18,28 @@
  */
 
 import { Router } from 'express';
+import { APIStatus } from 'common/src/service/Status';
```
anncatton, line 21: @mistryrn pointed this out to me, but we'll have to get these types to export correctly so that we can import nicely, i.e. `import { APIStatus } from 'common/service'` or even `import { APIStatus } from 'common'`. I was fiddling with the common build locally and haven't got it to work properly yet, so if necessary we can just add a `TODO` here to remind ourselves to fix, and not block the res of this work!

apps/consent-ui/src/constants/externalPaths.ts
```diff
@@ -0,0 +1,23 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// API
+export const API = {
```
anncatton, line 21: 👍

apps/consent-ui/src/constants/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+import { AxiosError, AxiosResponse } from 'axios';
+
+export interface APIResponse {
```
anncatton, line 21: I think specifically for `types` we're going to want to use `types.ts` files in the applicable directory, instead of putting them with constants. i'm undecided about how modular we want the types files to be but we can maybe decide as we get more things implemented  - or i @mistryrn has any thoughts/preferences for that?

apps/consent-ui/src/hooks/getAPIStatus.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import axios, { AxiosError, AxiosResponse } from 'axios';
+
+import { API, APIResponse } from '@/constants';
+import { getAppClientConfig } from '@/components/AppConfigContextProvider/utils';
+import getAppConfig from '@/getAppConfig';
+
+const axiosInstance = axios.create({
```
anncatton, line 26: how would you feel about a dedicated file we can export a base axios instance from, and then we can reuse as needed?

apps/consent-ui/src/hooks/getAPIStatus.tsx
```diff
@@ -0,0 +1,56 @@
...
+	validateStatus: (status: number) => status === 200,
+});
+
+const getAPIStatus = async () => {
```
anncatton, line 30: Apologies, I know I suggested earlier about creating a hook for this request, like the example in `dac-ui`! But looking at this now and where we call it, in a server component, i'm thinking we can simplify this somewhat and skip anything hooks-like. I'm thinking something like:
```
const getAPIStatus = async () => {
	const appClientConfig = await getAppClientConfig();
	const { CONSENT_API_URL } = getAppConfig(appClientConfig);

	return await axiosInstance
		.get(API.STATUS, { baseURL: CONSENT_API_URL })
		.then((res: AxiosResponse<APIStatus>) => {
			return res.data;
		})
		.catch((err: AxiosError<APIStatus>) => {
			console.error('Unable to receive consent-api status ⛔️: ', err);
			// we can decide on some custom error response, instead of just throwing an error and catching
			// so we have something to display in the case of an error
			const errorRes: APIStatus = {
				version: 'not found',
				status: 'API fetch failed',
			};
			return errorRes;
		});
};
```

And we can take advantage of the error and response types in Axios and customize with your `Status` type so that we can control the exact return type from this function.

anncatton: And then when the function is called in the `Versions` component, we can call it like:
`const { version } = await getAPIStatus()`
and maybe wrap the version `span` with `<Suspense fallback={<span />}>{...}</Suspense>`

apps/consent-ui/src/constants/index.ts
anncatton: I think also in this particular case we don't need this `APIResponse` type as we can reuse what Axios provides. One thing I think could be useful would be adding a specific "version error" type where we have a static response we can return from the request if an error occurs, i.e.:
`{ version: 'Unavailable', status: 'Version unavailable' }`
Just an example, we maybe just leave the API version blank if the req fails 🤔

anncatton (changes requested):
Thanks for updating the router! I've requested a few more changes for the fetching, but the type importing stuff is not necessary on this PR - we can address later if needed!

apps/consent-api/src/routers/health.ts
alyssadsouza: updated!

apps/consent-api/src/routers/status.ts
```diff
@@ -18,9 +18,10 @@
  */
 
 import { Router } from 'express';
-import { APIStatus } from 'common/src/service/Health';
+import { APIStatus } from 'common/src/service/Status';
 
-import { version } from '../../package.json';
+import { version } from '@/../package.json';
+import logger from '@/logger';
```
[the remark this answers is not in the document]
alyssadsouza: Sure! I'll do it in a separate PR.

apps/consent-api/src/routers/status.ts
```diff
@@ -18,9 +18,10 @@
  */
 
 import { Router } from 'express';
-import { APIStatus } from 'common/src/service/Health';
+import { APIStatus } from 'common/src/service/Status';
 
-import { version } from '../../package.json';
+import { version } from '@/../package.json';
```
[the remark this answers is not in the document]
alyssadsouza: I don't have a preference either, unless @mistryrn disagrees I'll leave it as relative

apps/consent-ui/src/hooks/getAPIStatus.tsx
alyssadsouza: I think it's a good idea since the base URL in the UI will always be the consent-api URL, so that's a good parameter to set in the base instance.

alyssadsouza: Although accessing the consent-api URL is asynchronous since we need the correct environment variable so that may be annoying to set.

apps/consent-api/src/routers/status.ts
alyssadsouza: added a TODO

apps/consent-ui/src/hooks/getAPIStatus.tsx
anncatton: Good call! Thanks for taking extra time with this. tbh i am still figuring out the whole server component approach so it's learning as we go 🙂

anncatton: one other thought here...since this function isn't a hook we should move it elsewhere. @mistryrn any thoughts on a nice way to structure this? i think it would also be good to move the `axios` instance creation out of `constants`, that's not really a fit either 🤔

anncatton (changes requested):
Thanks for all the fixes! Just a couple cleanup requests. And penny for yours and @mistryrn 's thoughts on how to structure the api request portion

apps/consent-ui/src/hooks/getAPIStatus.tsx
alyssadsouza: Maybe just a separate directory `src/api`? Though might be a bit confusing since there's `src/app/api` for the env variables stuff.

mistryrn (changes requested):
Sorry for taking so long to get through this 😅 This looks great!! Everything works as expected, and after spending **way** too long chatting with Ann about naming and organizing things, I've got a solution for the open conversations in the PR.

Apologies for making you move things around yet again, but this will be the last time for this PR (we promise)! 🙏 

Here's the new target structure:
```
.
└── consent-ui/
    └── src/
        ├── app
        ├── config/
        │   ├── appConfig.ts
        │   └── index.ts
        ├── constants/
        │   ├── externalPaths.ts
        │   └── index.ts
        └── services/
            └── api/
                ├── axiosClient.ts
                ├── status.tsx
                └── index.tsx
```

- [ ] move `getAppConfig.ts` to `/config/appConfig.ts`, create `/config/index.ts` to export everything so we can import from `@/config`, update imports
- [ ] move `constants/api.ts` to `services/api/axiosClient.ts`
- [x] move `hooks/getAPIStatus.tsx` to `services/api/status.tsx`, create `services/api/index.tsx` to export everything, update imports
- [ ] make Axios client in `services/api/axiosClient.ts` consent-api-specific (specify `baseUrl` in the config there, update requests that use it to no longer have to specify it in each request)

That should do it for this PR! Thanks so much for putting up with @anncatton and I!!

mistryrn (approved):
Looks great! 🚀🚀🚀

apps/consent-ui/src/services/api/status.ts
```diff
@@ -0,0 +1,45 @@
...
+const getAPIStatus = async () => {
+	const axiosClient = await getAxiosClient();
+	return await axiosClient
+		.get(API.STATUS)
+		.then((res: AxiosResponse<APIStatus>) => {
+			return res.data;
+		})
+		.catch((err: AxiosError<APIStatus>) => {
+			console.error('Unable to receive consent-api status ⛔️: ', err);
+			// we can decide on some custom error response, instead of just throwing an error and catching
```
anncatton, line 35: can delete my comments 🙂

anncatton (approved):
Looks great! You can delete my comment in the version fetch call as it was just for my example, but approved! 🎉 🚀

PR #89, opened 2023-09-13
Absolute paths fix

anncatton:


apps/consent-api/package.json
```diff
@@ -28,6 +28,7 @@
     "@types/swagger-ui-express": "^4.1.3",
     "dotenv": "^16.0.3",
     "eslint-config-ohcrn": "workspace:*",
-    "nodemon": "^2.0.22"
+    "nodemon": "^2.0.22",
+    "tsconfig-paths": "^4.2.0"
```
alyssadsouza, line 32: When I was doing some debugging the first time one of the suggestions I saw was this package. It's weird that it's not necessary for consent-ui even though both TS configurations are pretty much the same 🧐. Everything's working though!

anncatton: yeah i think nextjs is doing some magic for us behind the scenes!

PR #94, opened 2023-09-18
fix LocalizedLink href structure

anncatton:
adds a leading `/` to the `LocalizedLink` component href to fix a routing issue that (maybe?) appeared after the env var stuff got merged. TBH not sure how it was working originally!
- also fixes the prefetch link that can be seen with the browser inspector


mistryrn (approved):
Thanks for catching this! 🚀

Issue #58, opened 2023-09-07
Add Method for Fetching API Version


b-f-chan:
- [x] Add an Express endpoint to the `consent-api` `/health` router for fetching the API version # from the `package.json`, which will be displayed in the `consent-ui` Footer
- [x] Update the `consent-ui` Footer to consume this API and display the version number

Additionally, add some scaffolding to the `consent-ui` for grabbing the app config / env vars, and structure for using external routes:

- [x] Add a `config` dir, a `constants` dir, and a `hooks` dir to the `src/` directory of the `consent-ui`
- [x] Add an `index.ts` to the new `config` dir, similar to this example from dac-ui: https://github.com/icgc-argo/dac-ui/blob/develop/global/config/index.tsx
    - This `AppConfig` will be reused throughout the UI for getting config / env vars
- [x] Add an `externalPaths.ts` and an `index.ts` to the new `constants` dir, similar to this example from dac-ui: https://github.com/icgc-argo/dac-ui/blob/develop/global/constants/externalPaths.ts
    - These will be used for constants, in this case external path constants such as the `consent-api` URL and/or Consent API endpoints
- [x] Add a `useHealthAPI.tsx` hook to the new `hooks` dir, similar to this example from dac-ui: https://github.com/icgc-argo/dac-ui/blob/develop/global/hooks/useHealthAPI.tsx
    - This hook will be used to fetch values from the `consent-api` `/health` route endpoint

```
.
└── consent-ui/
    └── src/
        ├── config/
        │   └── index.ts
        ├── constants/
        │   ├── externalPaths.ts
        │   └── index.ts
        └── hooks/
            └── useHealthAPI.tsx
```

- [x] Add a shared type definition for the `Version` type returned from the `consent-api` to the `common` package
- [x] Update the structure of the `common` package as follows:
```
.
└── common/
    └── src/
        ├── service/
        │   └── Health/
        │       ├── Health.ts
        │       └── index.ts
        └── index.ts
```

b-f-chan:
reviewed, merged, closing

PR #98, opened 2023-09-19
💄📱 Implement landing page hero

alyssadsouza:
- Implemented hero with responsive styling
- Updated responsive styling for Footer to use the grid display so the sizing and arrangements are cleaner
- Added animations for Button component
- Added some Icon components for reusable SVGs

Screenshot 2023-09-19 at 5 09 52 PM
image
image


apps/consent-ui/src/components/Home/Card/Card.tsx
```diff
@@ -0,0 +1,75 @@
...
+import LocalizedLink from '@/components/Link/LocalizedLink';
+
+import styles from './Card.module.scss';
+import CardButton from './CardButton';
+
+const Card = async ({ currentLang }: { currentLang: ValidLanguage }) => {
+	const translate = await getTranslation(currentLang, 'landing-page');
+	return (
+		<div className={styles.card}>
+			<h3>{translate('join-ohcrn')}</h3>
```
mistryrn, line 35: One thing to keep in mind with headings is that for accessibility best practices, we should try not to skip heading levels. In this case (the landing page), it looks like the `h1` is "Welcome to the OHCRN Registry", and I believe this "Join OHCRN" should be an `h2` instead of an `h3`.

Not a big deal! Just makes things smoother for screen readers that take a look at the page hierarchy. If you want to learn more, here's a couple links to skim through if you're keen :) 
- https://www.a11yproject.com/posts/how-to-accessible-heading-structure/
- https://www.w3.org/WAI/tutorials/page-structure/headings/

apps/consent-ui/src/components/Home/Home.module.scss
mistryrn, line 1: Would it be possible to have the Footer not render below the fold if there's room? Not your fault, I think when we set up the scaffolding for the UI we made the Hero section fill the screen height on Desktop, but now that I see it in action I think it makes more sense to have it visible if the content height allows.

apps/consent-ui/tailwind.config.js
mistryrn, line 1: **TL;DR: we need to investigate why our tailwind responsive breakpoint overrides aren't working, and then ask Patrick what to do in the awkward zone between tablet and desktop**

Something we might want to consult @patrickdsantos on – after looking at the responsiveness in action, things feel a bit weird between our md/tablet breakpoint and our desktop designs (from around `1024px` wide to `1179px`)

It looks like our desktop designs are at a width of 1440px (90rem) in Figma, so maybe they should kick in after the `xl` breakpoint and we use the tablet designs for md and lg?

For reference, the default Tailwind breakpoints are:
- sm: 640px
- md: 768px
- lg: 1024px
- xl: 1280px
- 2xl: 1536px

And we modified `sm` and `md` in our tailwind config, resulting in:
- sm: 480px
- md: 1024px
- lg: 1024px
- xl: 1280px
- 2xl: 1536px

However, now that I look at it, maybe the modification hasn't worked? `md` CSS appears to be kicking in at 768px right now. Might require some investigation to see what's going on with our tailwind overrides. @anncatton any thoughts?

Once the breakpoints are working as expected, we might want to do a screenshare with Patrick to show him how things look at the different sizes so he can determine how to proceed in those "awkward" widths

apps/consent-ui/src/components/Home/Card/Card.tsx
```diff
@@ -0,0 +1,75 @@
...
+				</LocalizedLink>
+			</CardButton>
+			<CardButton Icon={DoctorIcon} classes={styles.mobileCardBtn}>
+				<LocalizedLink name={'invite'} linkLang={currentLang}>
+					{translate('clinicians-register-today')}
+					<RightArrow />
+				</LocalizedLink>
+			</CardButton>
+			{/* tablet/desktop view */}
+			<CardButton Icon={PatientIcon} classes={styles.tabletCardBtn}>
```
mistryrn, line 53: Something that might be nice from a usability perspective for the CardButton – can the entire block be a link? Rather than having to click specifically on the "Register yourself today" text?

apps/consent-ui/src/public/arrow.svg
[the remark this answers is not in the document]
alyssadsouza: I took a look at the designs for other pages and I do see the chevron arrow being used again. To be honest, I might have a preference for saving all the SVGs as components and not bother with saving them under `public` since they'd be automatically reusable and even if we don't customize them it still works as usual.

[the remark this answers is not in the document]
anncatton: i would like to take advantage of the optimizing that's automatically available when we use next/image, where possible...so for things like the corporate logos in the header/footer elements i'd like to leave as is, if you're all on board. but for the reusable icons like we'd have in buttons etc i'm happy with the React SVG approach!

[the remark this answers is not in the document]
alyssadsouza: makes sense!

apps/consent-ui/src/components/Home/Home.module.scss
alyssadsouza: I think the issue with that before was that the webpage might be too short if there's not enough content on one page, and sorting out the height of the main page content got a bit tricky with the header and the footer. I think I can fix this with relative viewport heights by setting a relative min height for the header, footer, and main page content, I'll update it for the large media breakpoint since that's the only viewport that has the issue.

apps/consent-ui/src/i18n/locales/en/footer.json
```diff
@@ -10,7 +10,7 @@
     "contact": "Contact",
     "privacy": "Privacy Policy",
     "terms": "Terms & Conditions",
-    "copyright": "Ontario Hereditary Cancer Research Network. All rights reserved.",
+    "copyright": "© 2022 Ontario Hereditary Cancer Research Network. All rights reserved.",
```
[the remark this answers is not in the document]
alyssadsouza: yup, updating now!

apps/consent-ui/src/components/Home/Card/Card.tsx
alyssadsouza: updated!

apps/consent-ui/src/components/Home/Card/Card.tsx
alyssadsouza: updated!

apps/consent-ui/tailwind.config.js
alyssadsouza: was able to figure out why the breakpoints weren't working - after checking the docs looks like it was just supposed to be named `screens` in the tailwind config rather than `breakpoints`! updated the responsive styling based on our current custom breakpoints for now.

anncatton (changes requested):
Looks good! Starting to look like a real website now 😁 Just a couple small changes requested. 
One thing i noticed when running it locally, the links in the footer seem to be left aligned on medium/small screens...i think on the responsive designs they were centered, but maybe that has changed since i first looked at them?

apps/consent-ui/src/components/Header/Header.module.scss
```diff
@@ -1,8 +1,8 @@
 .header {
-  height: 55px;
+  height: 3rem;
   @apply grid grid-cols-2;
   @media (min-width: theme('screens.md')) {
-    height: 80px;
+    min-height: 10vh;
```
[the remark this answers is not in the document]
alyssadsouza: Hmm I haven't actually heard anything about the header yet... I updated it so the header stays the same size (10vh i.e. 10% of the viewport height) but that's open to changes!

apps/consent-ui/src/components/Home/Card/Card.tsx
```diff
@@ -0,0 +1,89 @@
...
+	const translate = await getTranslation(currentLang, 'landing-page');
+	return (
+		<div className={styles.card}>
+			<h2>{translate('join-ohcrn')}</h2>
+			<p>
+				<b>{translate('join-ohcrn-description')}</b>
+			</p>
+			{/* mobile view */}
+			<CardLink
+				path={'register'}
```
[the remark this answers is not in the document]
alyssadsouza: I'll update to name since it's consistent with LocalizedLink 👍

apps/consent-ui/src/components/Home/Card/CardLink/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+	path,
+	currentLang,
+	Icon,
+	classes,
+	iconClasses,
+	children,
+}: {
+	path: RouteName;
+	currentLang: ValidLanguage;
+	Icon: any;
```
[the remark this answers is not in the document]
alyssadsouza: updated to `React.ComponentType<any>`

anncatton (approved):
Awesome work, thanks @alyssadsouza ! 🎉 🚀 
I'm good with merging this once the PR for #58 is done and merged in `develop` so we have separate merge commits for each ticket

mistryrn (approved):
Great stuff!! Thanks for getting those changes in so quickly (and fixing our breakpoints 😅 ) Checked out and tested locally, looks **awesome** – some really nice subtle hover animations and things to tie it all together 😄 

As Ann said, good to merge this once the PR for #58 gets merged in!

Issue #35, opened 2023-07-25
Configure Docker File for UI Service in QA Environment


b-f-chan:
Configure docker file for UI service in QA envt as part of cleanup tasks

May need to consult @mistryrn 

The results of this process will inform or be a template for the other platform services

**EXIT CRITERIA**:

- [x]  A docker file exists from which you can build the image of the UI service successfully

b-f-chan:
PR still in draft, pending review

b-f-chan:
Completed, closing

PR #100, opened 2023-09-21
🧱 Dockerfile Setup (#35)

mistryrn:
* Add `Dockerfile` for building `consent-ui` image
* Add `.dockerignore`
* Update `consent-ui` config to enable standalone build with flag
* Add `sharp` package to `consent-ui` for image optimization in build


Jenkinsfile.groovy
```diff
@@ -78,6 +86,80 @@ pipeline {
                 }
             }
         }
+
+        stage('Build') {
+            steps {
+                container('docker') {
+                    // the network=host needed to download dependencies using the host network (since we are inside 'docker' container)
```
justincorrigible, line 93:
```suggestion
                    // the network=host is needed to download dependencies using the host network (since we are inside 'docker' container)
```

apps/consent-ui/next.config.js
```diff
@@ -18,6 +18,8 @@ const nextConfig = {
 				})),
 		);
 	},
+	// use standalone build for Docker
+	output: process.env.BUILD_STANDALONE === 'true' ? 'standalone' : undefined,
```
justincorrigible, line 22: Not for this PR: for all env vars, in Arranger, I've been using [this formula](https://github.com/overture-stack/arranger/blob/develop/modules/server/src/utils/stringFns.ts#L2) in order to allow variations of 'true' that should be seen as valid (e.g. True, TRUE).

justincorrigible (approved):

.dockerignore
```diff
@@ -0,0 +1,18 @@
+# Deny All
+*
```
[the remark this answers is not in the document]
mistryrn: Good point, I guess it does make it more of a `.dockerallow` than a `.dockerignore` ;)

apps/consent-ui/next.config.js
mistryrn: Oh, I like that! Thanks!

justincorrigible (approved):

PR #101, opened 2023-09-25
🗃️ Add Participant Prisma schema to Consent DAS

alyssadsouza:
- Created the (updated) Participant Prisma schema with updated seed data and CRUD operations
- Also added and tested the router endpoints for Participant, Consent Question, and Participant Response CRUD operations
  - Added simple errorHandler (error-handler package) → just returns a 500 ServerError and generic message for now
  - Added basic Swagger JSDocs


apps/consent-das/src/routers/participantResponses.ts
```diff
@@ -0,0 +1,197 @@
...
+	try {
+		const participant_response = await createParticipantResponse({
+			participantId,
+			consentQuestionId,
+			response,
+		});
+		res.status(201).send({ participant_response });
+	} catch (error) {
+		logger.error(error);
+		if ((error as ErrorCallback).name == 'PrismaClientValidationError') {
```
joneubank, line 187: While the code here is going to work, from a strictly TS safe standpoint there are a couple issue here:

The type of `error` is `unknown`, so we need to do more validation upfront before interacting with it. if `error` is undefined (from a `throw undefined` which is not something we'll add but is theoretically possible) then `error.name` will break, throwing a server error from inside your catch block.

You can more thoroughly validate this with `if(error && 'name' in error && error.name === 'PrismaClientValidationError')`

Note this doesn't confirm for us that it has the other properties of a `PrismaClientValdiationError`, so you may also need to have a checker for those.

Yes this is cumbersome, this is a consequence of how JS lets you throw absolutely anything. This is why using throw logic is quite tricky overall :(

apps/consent-das/src/routers/participantResponses.ts
```diff
@@ -0,0 +1,197 @@
...
+ *       - bearerAuth: []
+ *     responses:
+ *       200:
+ *         description: The participant responses were successfully retrieved.
+ *       400:
+ *         description: Sort order specified was invalid.
+ *       500:
+ *         description: Could not retrieve participant responses.
+ */
+router.get('/all/:participantId/:consentQuestionId', async (req, res) => {
```
anncatton, line 56: i'm going to suggest for these 2 `GET` endpoints that we only need one with a sort option. the mapper/api layers can then have additional logic to either return the whole list, or the latest response
i also want to rename this path but i'm not exactly sure what makes the most sense, i will give it a little extra thought!

apps/consent-das/src/routers/participantResponses.ts
anncatton: After talking to @joneubank , let's add a super simple error handler on the router as a catch all for now, and we can discuss/implement custom (and typed) error handling in another ticket.

It can be added on the express app in the main `index.ts` file like `app.use(errorHandler);`
This is the example in the variant-indexer repo:
 https://github.com/EUCANCan-Federated-Data-Portal/variant-indexer/blob/main/src/middleware/errorHandler.ts

But all it needs to do for now is return a `500` error! We can leave in the try/catches so that structure is still there, but can remove the if/elses for the different error types.

anncatton (changes requested):
Lots of good stuff here @alyssadsouza ! Glad you're looking at error handling already 🙂 
A lot of my feedback is gonna involve separate prs (swagger, typed errors, absolute imports) but I think what we can do here is hone down the changes a little to known fixes, and we'll discuss the separate tasks with the team - esp the error handling as I think it'll be worthwhile to think about what types of errors we're dealing with at different layers of the backend.

apps/consent-das/src/routers/participantResponses.ts
alyssadsouza: so to clarify, we don't need an optional `submittedAt` parameter to get a response at the specified date anymore?

apps/consent-das/prisma/schema.prisma
```diff
@@ -8,11 +8,21 @@ generator client {
...
 model Participant {
-  id                  String                @id @default(cuid())
-  participantId       String                @unique
-  emailVerified       Boolean               @default(false)
-  ParticipantResponse ParticipantResponse[]
+  id                               String                @id @default(cuid())
+  consentGroup                     ConsentGroup?
+  emailVerified                    Boolean               @default(false)
+  registeringOnBehalfOfSomeoneElse Boolean
+  registrantIdVerified             Boolean?
```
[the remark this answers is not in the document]
anncatton: @mistryrn is on board with `isGuardian` so please proceed!

apps/consent-das/src/index.ts
```diff
@@ -22,13 +22,17 @@ import bodyParser from 'body-parser';
...
+import ConsentQuestionRouter from './routers/consentQuestions';
+import ParticipantResponseRouter from './routers/participantResponses';
 
 const App = (config: AppConfig) => {
 	const app = express();
 	app.set('port', config.port);
 	app.use(bodyParser.json());
 
 	app.use('/participants', ParticipantRouter);
+	app.use('/consent-questions', ConsentQuestionRouter);
```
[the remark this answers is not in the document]
anncatton: Just adding the [ticket link](https://github.com/OHCRN/platform/issues/116) here for reference on this PR

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -0,0 +1,208 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { Router } from 'express';
+import { ConsentCategory } from 'common';
+
+import { updateConsentQuestionIsActive } from '../service/update';
+import { getConsentQuestion, getConsentQuestions } from '../service/search';
+import { createConsentQuestion } from '../service/create';
+import logger from '../logger';
```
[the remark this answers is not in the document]
anncatton: Ticket created for this https://github.com/OHCRN/platform/issues/117

apps/consent-das/src/errorHandler.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Set default response for unhandled errors to be json instead of html.
+ * @param err
+ * @param _req
+ * @param res
+ * @param next
+ * @returns
+ */
+function errorHandler(err: unknown, _req: Request, res: Response, next: NextFunction): any {
```
anncatton, line 30: My bad, can you move this up to be a package utility like the `logger`? But no need to add in the other express routers yet, we can add it as those apps get worked on!

If you prefer, we can do this in a separate PR

anncatton (commented):
Just adding some links to additional tickets for reference on this pr.

For moving the error handler into its own package, I'm open to doing the change here or in a separate PR. Re-request a review when you're ready!

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -0,0 +1,208 @@
...
+ *     responses:
+ *       200:
+ *         description: The list of questions was successfully retrieved.
+ *       400:
+ *         description: The specified category was invalid.
+ *       500:
+ *         description: Could not fetch consent questions.
+ */
+router.get('/', async (req, res) => {
+	logger.info('GET /consent-questions');
```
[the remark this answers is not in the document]
mistryrn: @alyssadsouza sorry to jump in here late (and to conflict with what @anncatton suggested 😅 ) but for now can we leave these logs in and leave them at log level `INFO`?

When we start doing things end-to-end (especially once we get auth involved), we'll want logs on operations performed in the DASes for Auditing purposes. I agree that `INFO` probably isn't the right log level here, maybe we'll add an `AUDIT` level down the line, but for now let's leave these in and as `INFO` for consistency across the DASes and Data Mapper. It'll just make it easier to Shift+CMD+F and Find All when we get a clear picture of how we want to handle the Auditing

apps/consent-das/src/service/search.ts
```diff
@@ -6,11 +6,11 @@ import prisma, {
 	ConsentCategory,
 } from '../prismaClient';
 
-export const getParticipant = async (participantId: string): Promise<Participant> => {
+export const getParticipant = async (id: string): Promise<Participant> => {
```
mistryrn, line 9: ## TL;DR
Sorry to request another change @alyssadsouza , but after chatting with Ann I think for clarity we'd like to still use `entityId` when referring to an entity's `id`  in function calls. There's gonna be a lot of different IDs floating around the platform, this will just make it clear when making a function call exactly which ID you need to provide.

Do you mind updating all the functions and function calls to use `entityId` (i.e. `participantId` for Participant, `consentQuestionId` for ConsentQuestion, etc)?

## Full Context
I know we decided that inside the tables we’re using `id` as an entity’s ID (i.e. `consentQuestion` has `id` as it’s ID), and when referenced externally we refer to it as `entityId` (i.e. `consentQuestionId` as the foreign key repping a Consent Question ID).

This is pretty standard practice, makes sense, I’m all for that. However, when we’re dealing with functions that interact with an entity, I think I’d like to call the params `entityId` to avoid confusion.

For example, in the consent-das search service we currently have:
```typescript
export const getConsentQuestion = async (id: string): Promise<ConsentQuestion> => {
	const result = await prisma.consentQuestion.findUniqueOrThrow({
		where: {
			id,
		},
	});
	return result;
};
```

Nothing wrong with that functionally, but I think I’d prefer:
```typescript
export const getConsentQuestion = async (consentQuestionId: string): Promise<ConsentQuestion> => {
	const result = await prisma.consentQuestion.findUniqueOrThrow({
		where: {
			id: consentQuestionId,
		},
	});
	return result;
};
```

This way the VSCode TypeScript helper will show `getConsentQuestion(consentQuestionId: string)` and it’s very clear which ID to use when calling the function.

mistryrn (changes requested):
This is looking great! Noticed something I missed when reviewing the previous DAS PRs, sorry to request another change but this will improve the clarity of which ID is required when calling any of the DAS functions down the line 🙏

apps/consent-das/src/service/search.ts
alyssadsouza: makes sense! I'll update that now

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -159,7 +159,7 @@ router.post('/', async (req, res) => {
 	const { id, isActive, category } = req.body;
```
mistryrn, line 159: Can we also update the routers to expect the `id` as `entityId`? For example, with this route:

```suggestion
	const { consentQuestionId, isActive, category } = req.body;
```

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -159,7 +159,7 @@ router.post('/', async (req, res) => {
...
 	// TODO: add validation
 	try {
-		const question = await createConsentQuestion({ id, isActive, category });
+		const question = await createConsentQuestion({ consentQuestionId: id, isActive, category });
```
mistryrn, line 162: And then this line would become:

```suggestion
		const question = await createConsentQuestion({ consentQuestionId, isActive, category });
```

mistryrn: (I suppose the Postman collection might need updating after this too) 🙈

apps/consent-das/src/routers/consentQuestions.ts
alyssadsouza: do you want the request bodies updated too?

alyssadsouza: oh I'm blind haha never mind

anncatton (changes requested):
Sorry @alyssadsouza just a few more tidbits of feedback 😅 , and a couple thoughts for you and @mistryrn for future work on validation (that don't need answering here 😛 )

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -67,21 +67,20 @@ router.get('/', async (req, res) => {
 	if (category) {
 		try {
 			const questions = await getConsentQuestions(ConsentCategory.parse(category));
-			res.send({ questions });
+			res.status(200).send({ questions });
+		} catch (error) {
+			logger.error(error);
+			res.status(400).send({ error: 'Error retrieving consent questions' });
```
anncatton, line 73: Just change the 400 message to match the what you have in your jsdoc! 🙂

anncatton (approved):
Thanks for all this great work @alyssadsouza ! I'm approving, but let's make sure @mistryrn is all good with this before merging

mistryrn (approved):
🚀

PR #114, opened 2023-09-27
♻️ Refactor Consent UI to Fix Docker Build Errors

mistryrn:
## Summary
Some of our env var setup and network requests in `consent-ui` were causing build errors during docker builds due to various things (such as resources not being available at build time). This PR addresses those issues and refactors some of the env var setup and network requests.

Also grouped in are some README updates from a couple weeks ago that never got merged in 😅 

## Commits

♻️ Refactor Consent UI to Fix Docker Build Errors 
* Refactor runtime environment variable handling
* No longer expose env vars through an API endpoint
* Refactor Footer Consent API Version fetch to happen in a client component to prevent build errors
* Refactor shared Consent API axios client to create a single instance of axios
* Update missing translation check to prevent false negatives for empty strings
* Update Consent UI package.json scripts for `build` to specify `NEXT_IS_BUILDING` flag
* Add `consent-ui` env vars to root README

 📝 Add Root .env to README
* Add instructions on copying the root `.env.schema` into a `.env` for Docker Postgres setup


apps/consent-ui/src/components/Footer/APIVersion.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+'use client';
```
mistryrn, line 20: Moving this out into its own client component to prevent the Next.js build from over-optimizing and trying to fetch the API version at build time.

apps/consent-ui/src/components/Footer/APIVersion.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+	useEffect(() => {
+		const fetchData = async () => {
+			const { version } = await getAPIStatus();
+			setApiVersion(version);
+		};
+		fetchData();
+	}, []);
+
+	return <APIVersionLabel apiVersion={apiVersion} currentLang={currentLang} />;
```
mistryrn, line 40: Unfortunately, due to the whole Server Component vs Client Component kerfuffle, we end up having to compose this Footer "Versions" component in a very convoluted way because our translations are async.

In most cases, we can get around this by simply passing the translated content in as prop, however because this translation requires the API version (which must be fetched client-side), we've gotta nest the `APIVersionLabel` Server component inside this Client component 😂 Gotta love working around stupid over-optimizations.

apps/consent-ui/package.json
```diff
@@ -4,11 +4,11 @@
 	"private": true,
 	"scripts": {
 		"dev": "next dev",
-		"build": "next build",
+		"build": "NEXT_IS_BUILDING=true next build",
```
mistryrn, line 7: Add `NEXT_IS_BUILDING` flag to consent-ui build scripts to fix some Docker build issues

apps/consent-ui/src/app/api/status/route.ts
```diff
@@ -20,10 +20,12 @@
...
+/**
+ * GET /api/status
+ *
+ * Dummy Next.js local API route for the blocking data fetch in clientConfig.ts
+ */
 export async function GET() {
 	headers();
-	const appConfig = getAppConfig(process.env);
-	return NextResponse.json(appConfig);
+	return NextResponse.json('OK');
```
mistryrn, line 30: Converting this local Next.js API endpoint into a dummy endpoint since it didn't need to return any data. It was just meant to provide an endpoint to make a blocking fetch call to so that the `getClientAppConfig` function wouldn't get cached at build time (thus enabling us to expose runtime env vars to client components)

apps/consent-ui/src/services/api/axiosClient.ts
mistryrn, line 1: Updated the shared axios client to prevent creating a new client each time a request is made. We're now initializing a single axios client instance.

apps/consent-ui/src/config/appConfig.ts
```diff
@@ -42,15 +40,14 @@ const getAppConfig = (serverEnv: any): AppConfig => ({
 	 */
 	CONSENT_API_URL:
 		serverEnv.CONSENT_API_URL || process.env.CONSENT_API_URL || defaultAppConfig.CONSENT_API_URL,
-	CONSENT_URL: serverEnv.CONSENT_URL || process.env.CONSENT_URL || defaultAppConfig.CONSENT_URL,
+	CONSENT_UI_URL:
+		serverEnv.CONSENT_UI_URL || process.env.CONSENT_UI_URL || defaultAppConfig.CONSENT_UI_URL,
 	FEATURE_FLAG:
 		serverEnv.FEATURE_FLAG === 'false'
```
anncatton, line 46: this logic block is breaking my brain...i think we could simplify? but might help if i understand what it's doing (above question)

apps/consent-ui/src/components/Footer/APIVersion.tsx
anncatton: christopher nolan writes react code 🎩

anncatton (approved):
Thanks for righting some of my past wrongs @mistryrn 
Wondering if we can get rid of that route `participant/consent-forms` with the `ConsentButton` component, it looks like that's the only place where that `FEATURE_FLAG` var is used - maybe that's part of my confusion?

anncatton:
> @anncatton good call on removing the `FEATURE_FLAG` stuff! I'd like to leave it in until we get our first deploy to QA – it's the flag which was added to put that little 🔥 emoji on that consent button to test runtime environment variables working correctly.
> 
> Once that's tested in QA I'll make a commit to remove it!

Aha gotcha thanks for clarifying!

alyssadsouza (approved):
Thanks for updating the `axios` client function call! Looks good to me 👍

apps/consent-ui/src/components/Footer/APIVersion.tsx
mistryrn: @alyssadsouza unfortunately Next doesn't like it if you try to pass the server function into the client component as a prop 😞 I also tried using the translate function asynchronously in the `useEffect` but that produced some bizarre errors.

PR #119, opened 2023-09-28
🚑️ Fix renaming ButtonSize type bug

alyssadsouza:
- a dumb error that I didn't catch for some reason after renaming the types of ButtonSize (not sure why it was able to build without this 🤔)


anncatton (approved):
Good catch! yeah i wonder why ts didn't yell more aggressively 🤔 
Could be we're still missing some checks in the branch build

PR #121, opened 2023-10-02
🚨 Fix whitespace in Swagger JSDocs

alyssadsouza:
- didn't pull the swagger updates to develop before merging #101 so I wasn't able to see the jsdoc lib errors for the Swagger JSDocs
- couldn't find a programmatic way to fix this other than checking where the linter is complaining and removing tabs/adding spaces, swagger-jsdocs was being annoying when I tried to just return the swagger spec yaml or json
- I guess the current approach to avoiding this error is to check the linter as you add comments (which wasn't possible when I was adding these comments in originally so this should be fine from now on)
- also added NODE_ENV to consent-api's `.env.schema` and updated the README

### DIFF
--- a/apps/consent-das/src/routers/consentQuestions.ts
+++ b/apps/consent-das/src/routers/consentQuestions.ts
@@ -147,7 +147,12 @@ router.get('/:consentQuestionId', async (req, res) => {
  *                 type: boolean
  *               category:
  *                 type: string
- *                 enum: [INFORMED_CONSENT, CONSENT_RELEASE_DATA, CONSENT_RESEARCH_PARTICIPATION, CONSENT_RECONTACT, CONSENT_REVIEW_SIGN]
+ *                 enum:
+ *                   - INFORMED_CONSENT
+ *                   - CONSENT_RELEASE_DATA
+ *                   - CONSENT_RESEARCH_PARTICIPATION
+ *                   - CONSENT_RECONTACT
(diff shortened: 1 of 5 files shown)


apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -50,8 +50,13 @@ const router = Router();
  *         in: query
  *         description: Consent question category
  *         schema:
- *            type: Enum
- *            enum: [INFORMED_CONSENT, CONSENT_RELEASE_DATA, CONSENT_RESEARCH_PARTICIPATION, CONSENT_RECONTACT, CONSENT_REVIEW_SIGN]
+ *           type: string
```
samrichca, line 53: is there a benefit to these values being in a multi-line array versus a single line one? just curious because I haven't worked on these before

alyssadsouza: updating them all to be single lines because I think it's more readable for this many values

anncatton: yeah this is just one of the 100 ways to write a yaml, ugh. the hyphenated list is easier to read so i prefer that

mistryrn (approved):
Looks good! Thanks for fixing this 🙇

anncatton (approved):
thanks @alyssadsouza !

PR #124, opened 2023-10-05
Test bundler module res

anncatton:
Converting all apps and packages to use the ESM format, *EXCLUDING `consent-ui`*. This is the general direction JS is moving and we have already run into issues with newer versions of external libraries, i.e. `url-join` and `nanoid`.

Change summary:

**`package.json` files**:
- `type: module`
- enforce node version 18 or higher (with `.npmrc` `engine-strict=true` setting)
- set apps' "exports" to `null` as these shouldn't need to be imported anywhere
- switch `"main"` to `"exports"` field to match esm recommendation
- for all apps, `start` script will use `ts-node` to run from the `src` folder to avoid an issue encountered with the `prisma` generated client folder, which was not being added to the `dist` folder on build, as per @justincorrigible 's suggestion
- updates all `"workspace:^"` references to use the `^` instead of `*` syntax, for consistency. This is how it appears when installing a local package with `pnpm add <package>`

**`tsconfig.base.json`**:
- default `module` and `moduleResolution settings` will be `NodeNext` as recommended [here](https://github.com/microsoft/TypeScript/pull/54567), `target` becomes `ESNext`

**extended `tsconfig.json` files**:
- set `tsnode.esm` to true, cannot run without this
- set `tsnode.logError` to true to prevent apps running start script with `ts-node` from crashing if any type errors are encountered

**`consent-ui`**:
- overriding the base tsconfig with the `Bundler` option recommended [here](https://github.com/microsoft/TypeScript/pull/54567), to continue taking advantage of nextjs magic.
- not setting this app as `type: module` because this broke the postcss/tailwind styles
- removed the `@` alias for absolute imports, but will continue using this kind of import here

**General**:
- removes absolute paths as these do not work with esm modules.
- adds a `.js` file extensions to all imports, the path cannot be resolved properly otherwise
- adds version import from `package.json` in all apps. `assert { type: 'json' }` required on the import to align with changes in Node 17, as mentioned [here](https://stackoverflow.com/a/60206393)
- sets all package versions to `0.1.0`

**Packages**:
- `common` renamed to `types`
- restructured `types` folders
- modified `types` package.json to allow submodule imports, and updates all import statements

Documentation will be added for all of this in https://github.com/OHCRN/platform/issues/128


apps/consent-ui/src/components/Link/LocalizedLink.tsx
```diff
@@ -35,7 +36,7 @@ const LocalizedLink = ({
 	...rest
 }: LocalizedLinkProps) => {
 	let locale = linkLang;
```
anncatton, line 38: Haven't found a nice way to import these routes, so far only a json file has worked with the nextjs.config rewrites. Adding a `TODO` to look at a way to get rid of the `as` here

apps/consent-api/src/routers/status.ts
```diff
@@ -18,11 +18,10 @@
  */
 
 import { Router } from 'express';
-// TODO: fix exports from common (e.g. import { APIStatus } from 'common' or 'common/service')
-import { APIStatus } from 'common/src/service/Status';
+import { APIStatus } from 'types/services';
 
-import { version } from '@/../../package.json';
-import logger from '@/logger';
+import packageJson from '../../package.json' assert { type: 'json' };
```
justincorrigible, line 23: these absolute import paths can be troubleshot. I have a fix for them

apps/consent-das/prisma/seed/seed-data.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma, ConsentCategory } from '../../src/generated/client';
+import { Prisma, ConsentCategory } from '../../src/generated/client/index.js';
```
justincorrigible, line 1: oof. now this one "requirement" will sting for a lot of people 🤷

apps/consent-ui/src/app/[lang]/consent/step-1/page.tsx
```diff
@@ -17,10 +17,10 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { ConsentCategory } from 'common';
+import { ConsentCategory } from 'types/entities';
 
-import { ValidLanguage } from '@/i18n';
-import ConsentForm from '@/components/ConsentForm';
+import { ValidLanguage } from 'src/i18n';
```
Comment (justincorrigible) [inline apps/consent-ui/src/app/[lang]/consent/step-1/page.tsx:22]: does this absolute path work because of nextjs?

Comment (anncatton) [reply apps/consent-ui/src/app/[lang]/consent/step-1/page.tsx:22]: yes appears it is nextjs + the bundler setting magic

apps/consent-api/src/routers/status.ts
anncatton: @joneubank pointed out that absolute paths, while looking nice and tidy, can actually add confusion, and the esm standard is for relative paths anyway

apps/consent-das/prisma/seed/seed-data.ts
anncatton: yuuuuup.

apps/consent-api/src/routers/status.ts
joneubank: they can also cleanup in some cases. ideally we could use both.

justincorrigible: your right, relative is the ESM standard, but they make PR reviewing (as you don't have a full IDE in github), and they also make app onboarding so much harder (chasing the "dot dot rabbit" down its hole, vs "ah, this is from the global utils").

besides, just like having types in a js file (which is not js spec either), absolute paths are something we can handle through TS transpilation

packages/types/package.json
```diff
@@ -10,8 +10,8 @@
   },
   "type": "module",
   "engines": {
-		"node": ">=18"
-	},
+    "node": ">=18"
+  },
```
justincorrigible, line 14: curious about this change. is this project using tabs or spaces?

apps/data-mapper/tsconfig.json
```diff
@@ -5,5 +5,9 @@
     "rootDir": "src",
     "baseUrl": "."
   },
+  "ts-node": {
+    "esm": true,
+  },
   "include": ["src"],
-}
+  "exclude": ["node_modules"]
+}
```
justincorrigible, line 13:
```suggestion
}

```

packages/types/src/entities/ConsentCategory.ts
```diff
@@ -28,4 +20,4 @@ const CONSENT_GROUPS = [
 ] as const;
 
 export const ConsentGroup = z.enum(CONSENT_GROUPS);
-export type ConsentGroup = z.infer<typeof ConsentGroup>;
+export type ConsentGroup = z.infer<typeof ConsentGroup>;
```
justincorrigible, line 23:
```suggestion
export type ConsentGroup = z.infer<typeof ConsentGroup>;

```

packages/types/src/entities/User.ts
```diff
@@ -0,0 +1,9 @@
+import * as z from 'zod';
+
+export const User = z.object({
+	id: z.string(),
+	name: z.string(),
+	email: z.string().email(),
+});
+
+export type User = z.infer<typeof User>;
```
justincorrigible, line 9:
```suggestion
export type User = z.infer<typeof User>;

```

packages/types/src/entities/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './User.js';
+export * from './ConsentCategory.js';
```
justincorrigible, line 2:
```suggestion
export * from './ConsentCategory.js';

```

justincorrigible (approved):

packages/types/package.json
mistryrn: I think the linter settings (which apply things like tabs instead of spaces on fix-on-save) haven't been configured for all files in the repo (like package.json files), and when I was initially setting up the repo I was still using spaces so this may be a holdover from back then

In other words... my bad 🙈 This one's totally on me haha

justincorrigible: absolutely terrible. I'm heartbroken!!
how dare you!? 😂

/s

justincorrigible (dismissed):

mistryrn (approved):
Looks good! Tested locally and confirmed things are still building, running in dev, and now *actually* running with the `start` command too 😅 Thanks so much for going down the rabbit holes required to get this all in! 🙌 🙇 

Other than the little README update we talked about on Slack, this looks good to go 🚀

README.md
```diff
@@ -41,12 +41,14 @@ The directory structure is as follows:
 │       └── src
 ├── docker-scripts/                 ← Docker Init Scripts
 └── packages/
-    ├── common/                     ← Shared Validation & Types
-    │   └── src
+    ├── types/                      ← Shared Validation & Types/
```
mistryrn, line 44: nitpick: Looks like we've got some rogue trailing slashes here: `& Types/`, `Logger/`, `APIs/` -- also do you mind putting these in alphabetical order like they appear in the directory?

mistryrn (dismissed):
🚀

mistryrn (approved):
🚀  (for reals this time)

PR #125, opened 2023-10-05
🗃️ Add Participant Prisma schema to PI DAS

alyssadsouza:
- added the `Participant` Prisma schema and seed data
- added create + read + (general) update operations
- added the `/participants` router + swagger jsdocs


apps/pi-das/prisma/migrations/20230927144655_participant/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+-- CreateTable
+CREATE TABLE "Participant" (
+  "id" TEXT NOT NULL,
+  "dateOfBirth" TIMESTAMP(3) NOT NULL,
+  "emailAddress" TEXT NOT NULL,
+  "guardianEmailAddress" TEXT,
+  "guardianName" TEXT,
+  "guardianPhoneNumber" CHAR(10),
+  "guardianRelationship" TEXT,
+  "inviteId" TEXT,
```
alyssadsouza, line 13: note that this isn't documented in the Data Model tables on the Wiki, but is included in v3 of Hardeep's spreadsheet as optional so I've added it here. the foreign key constraint is added in #126

apps/pi-das/prisma/migrations/20230927144655_participant/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+  "mailingAddressCity" TEXT,
+  "mailingAddressPostalCode" CHAR(6),
+  "mailingAddressProvince" TEXT,
```
alyssadsouza, line 16: will update this to be a dropdown (Enum type)

apps/pi-das/prisma/migrations/20230927144655_participant/migration.sql
```diff
@@ -0,0 +1,29 @@
+-- DropTable
+DROP TABLE "Participant";
+
+-- CreateTable
+CREATE TABLE "Participant" (
+  "id" TEXT NOT NULL,
+  "dateOfBirth" TIMESTAMP(3) NOT NULL,
```
alyssadsouza, line 7: considering updating from `TIMESTAMP(3)` to `DATE` (by [adding `@db.Date` property](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#postgresql-6) in Prisma schema)

mistryrn (approved):
Looks good! Nice details in the seed data 😂 Thanks for updating for the ESM stuff too!

Issue #79, opened 2023-09-12
[Create PRISMA Schema] PHI Database > Clinical Profile


b-f-chan:
**EXIT CRITERIA:**

- [x] PRISMA schema exists for the this table within the PHI DAS
- [x] Provide seed data for this table
- [x] Basic CRUD operations exist in the PHI DAS

b-f-chan:
Merged, closed

PR #129, opened 2023-10-06
🗃️ #79: add clinical profile schema to PHI-DAS

samrichca:
- add clinical profile schema to PHI-DAS


apps/phi-das/prisma/seed/seed-data.ts
```diff
@@ -1,23 +1,83 @@
...
 		id: 'cllgsrruu000208mnhq6wh7e6',
 		ohipPrivateKey: 'cllgsrw92000308mn09obgujg',
-		ohipNumber: '1234567890CC'
+		ohipNumber: '1234567890CC',
+	},
+];
+
+const clinicalProfiles: Prisma.ClinicalProfileCreateInput[] = [
+	{
+		// both gender fields skipped
```
mistryrn, line 49: Thanks for adding these descriptions to the seed data!

mistryrn (dismissed):
Looks good!

alyssadsouza (changes requested):
Looks good! Could you add the labels for the enums in the `consent-ui` too? 🙏

apps/phi-das/prisma/schema.prisma
```diff
@@ -5,11 +5,330 @@ datasource db {
...
+  TRANSGENDER_MAN_TRANSMAN
+  TRANSGENDER_WOMAN_TRANSWOMAN
+  TWO_SPIRIT
+  WOMAN
+}
+
+enum HistoryOfCancer {
+  YES
+  NO
+  UNKNOWN
```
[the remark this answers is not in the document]
samrichca: this I am not sure about 🤔

apps/phi-das/prisma/schema.prisma
```diff
@@ -5,11 +5,330 @@ datasource db {
...
+}
+
+enum Ancestry {
```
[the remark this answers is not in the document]
samrichca: don't be sorry at all, this is a good idea! I'll add them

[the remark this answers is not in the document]
samrichca: excel/google sheets does string concatenations btw 🤓

apps/phi-das/prisma/schema.prisma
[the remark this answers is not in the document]
samrichca: resolved on slack, keeping as an enum

apps/consent-ui/src/i18n/locales/en/common.json
mistryrn, line 1: Thanks for adding the i18n labels for these! Now that I see it all in one file, I feel like it may be a good time to part out some of the `common` namespace before it gets too big.

Maybe we should create namespaces (JSON file + updating line 22 in `i18n/types/dictionaries.ts`) corresponding to the enums? That would take us from just `common`, to:
- common
- consentCategory
- consentGroup
- province
- ancestry
- ...etc

Open to suggestions, I'm not sure that's the right solution here but I am concerned with the growing size of the `common.json` files

mistryrn (commented):
Looks good! Just one comment now that I see how big the `common.json` file is getting.

Fortunately Ann is back on Monday and I know she had some thoughts on the i18n locale organization stuff. I'd want to hear her thoughts before splitting up that common file, just in case she was planning something already 🙏

apps/consent-ui/src/i18n/locales/en/common.json
anncatton: thanks for bringing this up @mistryrn ! you're right, the `common` file was just a catchall for when i was doing the POC work for translations. i'll take a quick gander at what we've got in there now compared with the other dictionaries and see what might make sense for organizing going forward.

anncatton: So my thought here is to organize these based on the components they’re used in, so that we can import specific dictionaries in specific places - i.e. like we have the `footer` dictionary imported in the `Footer` component, `header` dictionary in `Header` etc. We also have the option to import multiple dictionaries in one component, such as
 
```
const translateFromDictionaryOne = await getTranslation(currentLang, 'dictionary-one');
const translateFromDictionaryTwo = await getTranslation(currentLang, 'dictionary-two');
```
But i think the goal can be to keep this to a minimum, and maybe this approach can be what guides us in what dictionary files we create. So we could start adding dictionaries per page/component:

`clinician-registration`
`consent-step-1`
`consent-step-2`
`dashboard`

and keep `footer` , `header`, `landing-page` and so on.

I do agree with @mistryrn 's suggestion to split out some of these enums into their own dictionaries just because they'll get unwieldy, even if they are used in just one component. And they could still be imported as in the above multiple import example.

I imagine there will be some common bits of text that will be reused on various pages/components, and perhaps those are items we can put in a `common` dictionary, but we can map those out as we come to them. Also keep in mind some of the values i added previously (for page titles etc) were just for demo purposes, they may not be needed.

Another change I'd like to make - and not for this pr, it's just a task i have in mind that should be handled separately: i want to convert these dictionary files to `.ts` so we can get some type safety, and we'll also then be able to use our shared enum types as dictionary keys (and maybe some other types??), like:

```
[ConsentGroup.enum.ADULT_CONSENT]: 'Adult Consent (>18)'
```

apps/phi-das/src/routers/clinicalProfiles.ts
```diff
@@ -0,0 +1,176 @@
...
+ *       required: true
+ *       content:
+ *         application/json:
+ *           schema:
+ *             type: object
+ *             properties:
+ *               participantId:
+ *                 type: string
+ *                 required: true
+ *               birthSex:
```
anncatton, line 120: can we add the enum values for all the enum fields? To make it tidier we can add them as component schemas. i made an example pr here: https://github.com/OHCRN/platform/pull/158

swagger-jsdocs will pick up any files specified in the `api` list in the config options, however it doesn't seem to like folders so i've narrowed the paths to have file extensions - this also gives the option have multiple schema def files. as long as the yaml structure uses a  `components/schemas/ComponentName` setup it will be picked up by the `$ref`. Note that if there are any naming collisions between definition files swagger won't complain, the last defined component will just override! i think most of the enums schema defs could go in one file, but maybe we can separate `Ancestry` into its own file since it's gargantuan.

Also i tried this with json instead of yaml and got some sort of json pointer error...not sure if it was my code or jsdocs just doesn't like it...but i didn't feel like arguing with it so 😛

anncatton (changes requested):
Just added an additional request on the jsdocs stuff, let me know if you have any questions/thoughts on the sample pr!

alyssadsouza (dismissed):
Looks good to me!

mistryrn (dismissed):

apps/phi-das/src/routers/swagger-schemas/definitions.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - ORILLIA_SOLDIERS_MEMORIAL_HOSPITAL_ORILLIA_SIMCOE_MUSKOKA
+        - PETERBOROUGH_REGIONAL_HEALTH_CENTRE_PETERBOROUGH
+        - PRINCESS_MARGARET_HOSPITAL_TORONTO
+        - SINAI_HEALTH_SYSTEM_TORONTO
+        - SUNNYBROOK_HEALTH_SCIENCES_TORONTO
+        - THUNDER_BAY_REGIONAL_HEALTH_SCIENCES_CENTRE_THUNDER_BAY
+        - TRILLIUM_HEALTH_PARTNERS_MISSISSAUGA
+        - WINDSOR_REGIONAL_HOSPITAL_WINDSOR
+        - WOMENS_COLLEGE_HOSPITAL_TORONTO
+    FamilyHistoryOfCancer:
```
anncatton, line 49: Not necessary atm, but might be useful to create a more generic enum type for the `YES / NO / UNKNOWN` field types

apps/phi-das/src/routers/clinicalProfiles.ts
```diff
@@ -0,0 +1,182 @@
...
+ *                 type: string
+ *                 required: true
+ *                 $ref: '#/components/schemas/GeneticsClinic'
+ *               historyOfCancer:
+ *                 type: boolean
+ *                 required: true
+ *                 $ref: '#/components/schemas/HistoryOfCancer'
+ *               familyHistoryOfCancer:
+ *                 type: boolean
+ *                 required: true
```
anncatton, line 143: You can remove the `type` and `required` definitions here since that's defined in the schemas

anncatton (commented):
Thanks for adding the swagger schemas! Just one smol change needed there and then should be good to go

anncatton (approved):
woohoo, looks good to go 🚀

Issue #99, opened 2023-09-20
Footer, Header elements should be sticky


anncatton:
Update Footer and Header elements to `sticky`. The footer will only be sticky if there is minimal content, see this for a reference on the recommended approach:
https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers

b-f-chan:
@b-f-chan @patrickdsantos to test on QA:
https://consent-ui.dev.ohcrn.cancercollaboratory.org/en/register

patrickdsantos:
Footer seems good to me! But the header doesn't appear at the top when the window is full screen.

b-f-chan:
@samrichca can this be closed?

PR #132, opened 2023-10-06
💄#99: make header & footer sticky

samrichca:
- add position sticky to header
- use css flexbox to make footer stick to bottom of the page when the page is short

### DIFF
--- a/apps/consent-ui/src/components/Header/Header.module.scss
+++ b/apps/consent-ui/src/components/Header/Header.module.scss
@@ -1,64 +1,67 @@
 .header {
-  min-height: 10vh;
-  @apply grid grid-cols-2;
-  background-color: theme('colors.grayscale.100');
-  padding: 0.5rem 1rem;
-  align-items: center;
+	min-height: 10vh;
+	@apply grid grid-cols-2;
+	background-color: theme('colors.grayscale.100');
+	padding: 0.5rem 1rem;
+	align-items: center;
+	position: sticky;
+	top: 0;
+	left: 0;
+	right: 0;
 }
 
 .headerItem {
-  margin-right: 1rem;
+	margin-right: 1rem;
 }
 
 .right {
-  display: flex;
-  flex-direction: row;
-  justify-content: flex-end;
-  align-items: center;
+	display: flex;
+	flex-direction: row;
+	justify-content: flex-end;
+	align-items: center;
 }
 
 .help {
-  @apply hidden;
-  @media (min-width: theme('screens.md')) {
-    @apply flex;
-    margin-right: 1rem;
-  }
-  .helpButton {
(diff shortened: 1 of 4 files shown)


apps/consent-ui/src/components/Header/Header.module.scss
```diff
@@ -1,64 +1,67 @@
...
-  @apply grid grid-cols-2;
-  background-color: theme('colors.grayscale.100');
-  padding: 0.5rem 1rem;
-  align-items: center;
+	min-height: 10vh;
+	@apply grid grid-cols-2;
+	background-color: theme('colors.grayscale.100');
+	padding: 0.5rem 1rem;
+	align-items: center;
+	position: sticky;
```
alyssadsouza, line 7: tested this out and it works great! only thing is for the landing page you might need to set `z-index: 1` so that the header is always visible, I think this might have to do with how I have the gradient overlayed in the landing page hero.

apps/consent-ui/src/components/Header/Header.module.scss
```diff
@@ -1,64 +1,67 @@
 .header {
-  min-height: 10vh;
```
alyssadsouza, line 2: since you made the header sticky, we don't need this `min-height` because it'll always remain the same height based on its content (for context: originally when I implemented the landing page hero I was trying to fix the page being too short with `vh` sizing but I like this sticky header/footer approach better!) so you can get rid of this, and same for the footer `min-height`

alyssadsouza (changes requested):
Thanks for doing this! I left a couple small comments because I noticed the responsive views (mobile viewport width in particular) for the landing page needed a fix.

apps/consent-ui/src/components/Header/Header.module.scss
samrichca: thanks for the feedback - updated my PR

alyssadsouza: Thanks for updating it! Just one more thing, I know this wasn't in any of the files you changed but could you remove the `min-height: 20vh` style in the [`screens.md` viewport](https://github.com/OHCRN/platform/blob/eb03cc53576cac26b296b8cfcb412ab1c3906b24/apps/consent-ui/src/components/Footer/Footer.module.scss#L12C1-L12C1) (again, leftover from when I was using `vh` to fix the page sizing). It's a wonky dimension to catch but it messes up the footer here:

with `min-height` set:
Screenshot 2023-10-12 at 12 21 07 PM

without:
Screenshot 2023-10-12 at 12 21 15 PM

alyssadsouza (changes requested):

alyssadsouza (approved):
Looks good to me, thanks for this! 🚀

Issue #95, opened 2023-09-18
Create Notification Message UI Component


b-f-chan:
Create the notification message UI component to be used throughout the app; this is the component only, there is no logic or triggers tied to this

Check with @patrickdsantos style guide for styling

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdG1zIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e169899c99f0400727db6dc2d8732d48b5799292/image.png)

**Exit Criteria:**

- [ ] Generic UI component exists for the notification message that can be reused in any page in the portal and meets the responsive UX requirements in the style guide
- [ ] Component must have ability to have multiple messages displayed simultaneously

samrichca:
things not covered by this ticket:
- methods for showing/hiding notifications
- persisting the user's choice to dismiss a notification

b-f-chan:
Confirmed with @patrickdsantos the screenshots are sufficient, closing

PR #134, opened 2023-10-11
💄 ✨#95: notifications for consent UI (develop branch version)

samrichca:


packages/common/src/utils/tsUtils.ts
mistryrn, line 1: What was this file for?

mistryrn (approved):

PR #144, opened 2023-10-11
🦺 Add ParticipantIdentification Zod schema

alyssadsouza:
- Added the ParticipantIdentification Zod schema
- Moved the ConsentGroup enum and type to its own entity
- Created additional entities for common regex and Zod schemas/enums:
    - `Regex.ts`
    - `Name.ts`
    - `NanoId.ts`
    - `OhipNumber.ts`
    - `PhoneNumber.ts`
    - `PostalCode.ts`
    - `Province.ts`
- Updated the `zod` import statements based on their docs and added copyrights to all the entities
- Added simple tests in the common `types/` directory for testing custom Zod schema validation (i.e. regex and conditionally required fields)
  - `pnpm test` to run tests inside `packages/types` directory

Some notes:
- no further validation on the `emailAddress` field beyond the basic `email()` Zod validation since we have email verification for participants
- didn't add the `length(x)` Zod restriction on relevant fields because the `regex()` is more specific about the number of character occurrences
- `dateOfBirth` is a [`date()`](https://zod.dev/?id=dates) field as opposed to the [`string().datetime()`](https://zod.dev/?id=iso-datetimes) field because as noted in their docs, Prisma `DateTime` fields are always returned as `Date()` objects: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#remarks-3
- prefer `optional()` to `nullable()` because `undefined` (instead of `null`) should be expected everywhere except for Prisma client data


packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,62 @@
+import * as z from 'zod';
+
+import { ConsentGroup } from './ConsentCategory.js';
+
+const PROVINCES = [
```
mistryrn, line 5: Bit of an organization question – since this enum will probably end up being re-used for more than just Participant identification, do you think we should spin this plus the corresponding z.enum and type into its own `entities/Province.ts` file? Open to suggestions, not entirely sure what the best approach is

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,62 @@
...
+	'QUEBEC',
+	'SASKATCHEWAN',
+	'YUKON',
+] as const;
+
+export const Province = z.enum(PROVINCES);
+export type Province = z.infer<typeof Province>;
+
+export const ParticipantIdentification = z.object({
+	id: z.string().cuid(),
```
mistryrn, line 25: Something we may want to flag with `TODO`s – Ann will be changing our IDs to `nanoid` instead of `cuid` next week when she gets back.

I don't think Zod has a built-in `nanoid` string check like it does with `cuid` though. But we are configuring our nanoid setup with some specific params (length, characters) so we can come up with an appropriate regex.

Something like `// TODO: change from .cuid() to nanoid regex once nanoid is implemented` should be enough to track the work here

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,62 @@
...
+	participantPreferredName: z.string().regex(/^[A-Za-z\s]+$/),
+	participantOhipFirstName: z.string().regex(/^[A-Za-z\s]+$/),
+	participantOhipLastName: z.string().regex(/^[A-Za-z\s]+$/),
+	participantOhipMiddleName: z
+		.string()
+		.regex(/^[A-Za-z\s]+$/)
+		.optional(),
+	dateOfBirth: z.date(),
+	phoneNumber: z.string().regex(/[0-9]{10}/),
+	mailingAddressStreet: z.string().optional(),
```
mistryrn, line 37: Something to look into: I'm not sure how it'll interact with our existing Prisma setup for optional fields, but Zod also has `.nullable()` https://zod.dev/?id=nullables

Depending on how we implement updates to a participant (i.e. if someone wants to clear data from an optional field and they delete the string from the form field and then submit, does that send a `null` value?) we may need to make these nullable as well

Jon may be able to provide better insight into this

packages/types/src/entities/ParticipantIdentification.ts
alyssadsouza: Sure that's probably a good idea, I was considering it but decided not to since it's not used anywhere else in the current data models. That being said, something like Provinces is likely to be reused so I'll move it to its own file.

packages/types/src/entities/ParticipantIdentification.ts
alyssadsouza: Assuming this schema is supposed to be as close as possible to the data the backend retrieves, then it will need to send a `null` value to the Prisma client to delete that value (and subsequently returns a `null` value for that field if it has been deleted or not set), so `nullable()` makes sense. I think it's also fine to use `nullable()` here because we can chain `optional()` to any derivative schemas that are used for things like API request bodies that should accept undefined values too (since the Prisma client will read that as no changes needed for those fields in update operations).

alyssadsouza: Also for form submissions in the UI, if the user deletes information from a form and it gets set to an empty string, we can always modify the Zod schema for that form input to parse it as `null` rather than `""`:

```
z.string().transform((data) => data === "" ? null : data);
```

https://zod.dev/?id=transform

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,46 @@
...
+	mailingAddressCity: z.string().optional(),
+	mailingAddressProvince: Province.optional(),
+	mailingAddressPostalCode: z
+		.string()
+		.regex(/^[A-Za-z][0-9][A-Za-z][0-9][A-Za-z][0-9]/)
+		.optional(),
+	residentialPostalCode: z.string().regex(/^[A-Za-z][0-9][A-Za-z][0-9][A-Za-z][0-9]/),
```
samrichca, line 28: having a regex for canadian postal codes stored as a constant would be very helpful

alyssadsouza: I'm wondering if a Zod object for postal codes and one for phone numbers would be a good idea 🤔

packages/types/src/entities/ParticipantIdentification.ts
joneubank: We can create a custom nanoid validator: https://zod.dev/?id=custom-schemas

joneubank: Something like:
`const zodNanoid = zod.string().regex(/^[A-Za-z0-9_-]{21,36}$/);`

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,46 @@
+import * as z from 'zod';
```
joneubank, line 1: The recommended import from the zod docs is `import { z } from 'zod';`

@mistryrn I'm not sure if you have an opinion on this but i don't like single letter variable names, so I have been using this with:

`import { z as zod } from 'zod';`

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { ConsentGroup } from './ConsentGroup.js';
+import { Province } from './Province.js';
+
+export const ParticipantIdentification = z.object({
+	// TODO: change from .cuid() to nanoid regex once nanoid is implemented
+	id: z.string().cuid(),
+	// TODO: change from .cuid() to nanoid regex once nanoid is implemented
+	inviteId: z.string().cuid().optional(),
+	ohipNumber: z.string().regex(/[0-9]{10}/),
```
joneubank, line 11: We should probably have a dedicated OHIPNumber schema, or at least have the regex be a common export.

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,46 @@
...
+	participantPreferredName: z.string().regex(/^[A-Za-z\s]+$/),
```
joneubank, line 12: Repeated name regex can be a constant we reuse

packages/types/src/entities/ParticipantIdentification.ts
joneubank: Please don't introduce nulls into our types. `undefined` is the proper way to handle optional fields in JSON objects. The data going into and out of postgres can include nulls, but lets enforce undefined and not null elsewhere.

joneubank: (ie. `optional` is correct here)

packages/types/src/entities/ParticipantIdentification.ts
joneubank: yes please!

joneubank: Maybe we also want to have a transform on postal codes to enforce upper case?

joneubank:
This is great first step! Left a couple comments, mostly on reducing repeated regex statements.

Nitpicking on the name, perhaps it should be `ParticipantIdentity`?

samrichca (dismissed):
lgtm! 🚀 :shipit:

mistryrn (dismissed):

samrichca (approved):
refinement looks good 👍

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -0,0 +1,67 @@
...
+		const validateGuardianInformation = !(
+			(input.consentGroup === ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR ||
+				input.consentGroup === ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT) &&
+			(input.guardianName === undefined ||
+				input.guardianPhoneNumber === undefined ||
+				input.guardianEmailAddress === undefined ||
+				input.guardianRelationship === undefined)
+		);
+		return validateGuardianInformation;
+	});
```
joneubank, line 65: Thank you for the clear comment to describe why the refine block is here. This is so much easier to read to quickly understand the purpose of the code than having to parse out what the boolean logic is doing. Can also validate that the code is doing what the comment intends it to.

That said, I might try to simplify this logic as this is a mess of boolean properties.

First, I would consider separating the condition when this logic is applied into its own property:

```ts
const requiresGuardianInformation = 
    input.consentGroup === ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR ||
    input.consentGroup === ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT);
```

We don't want the entire returned condition to use a negation, that is quite a bit of extra workload. You can do this as a ternary or an if block:

```ts
// option 1 as ternary
return requiresGuardianInformation 
    ? input.guardianName !== undefined 
       && input.guardianPhoneNumber !== undefined 
       && input.guardianEmailAddress !== undefined 
       && input.guardianRelationship !== undefined
    : true;
```

packages/types/tests/entities.spec.ts
```diff
@@ -0,0 +1,140 @@
+import { expect } from 'chai';
```
samrichca, line 1: good idea! thanks for adding these

mistryrn (approved):

PR #147, opened 2023-10-12
🦺 Add ClinicianInvite Zod schema

alyssadsouza:
- Added the ClinicianInvite Zod schema
- Abstracted the conditionally required guardian contact fields refine logic to a reusable function
- Added unit tests for ClinicianInvite
- Also updated the `inviteSentDate` field in the Prisma schema to be required based on Hardeep's data model

There are a few other fields that were included in the data model table on the Wiki but not in Hardeep's data model and thus were added (because they were added in the Prisma schema):
- inviteAcceptedDate
- inviteAccepted
- participantEmailAddress
- participantPhoneNumber


mistryrn (approved):

packages/types/src/entities/ClinicianInvite.ts
```diff
@@ -0,0 +1,62 @@
...
+		consentGroup: ConsentGroup,
+		guardianName: Name.optional(),
+		guardianPhoneNumber: PhoneNumber.optional(),
+		guardianEmailAddress: z.string().email().optional(),
+		guardianRelationship: Name.optional(),
+	})
+	.refine((input) => {
+		// guardianName, guardianPhoneNumber, guardianEmailAddress, guardianRelationship must be defined if
+		// ConsentGroup.GUARDIAN_CONSENT_OF_MINOR or ConsentGroup.GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT was selected
+		const validateGuardianInformation = !(
```
samrichca, line 51: Jon had a good point about breaking this down into 2 conditions, should do that here too

packages/types/tests/entities.spec.ts
```diff
@@ -140,3 +141,59 @@ describe('ParticipantIdentification', () => {
 		).false;
 	});
 });
+
+describe('ClinicianInvite', () => {
```
samrichca, line 145: please move this to tests/entities/ClinicianInvite.spec.ts (i moved the other tests on my current working branch)

lots more entities to go, this file is going to be huge

packages/types/src/entities/ClinicianInvite.ts
```diff
@@ -0,0 +1,66 @@
...
+import { PhoneNumber } from './PhoneNumber.js';
+import { NanoId } from './NanoId.js';
+import { hasRequiredGuardianInformation } from './ParticipantIdentification.js';
+
+export const ClinicianInvite = z
+	.object({
+		id: NanoId,
+		inviteSentDate: z.date(),
+		inviteAcceptedDate: z.date().optional(),
+		inviteAccepted: z.boolean().optional(),
```
anncatton, line 33: Just curious here if defaulting to `false` would work, instead of an optional value?

packages/types/src/entities/ClinicianInvite.ts
```diff
@@ -0,0 +1,66 @@
...
+		clinicianFirstName: Name,
+		clinicianLastName: Name,
+		clinicianInstitutionalEmailAddress: z.string().email(),
+		clinicianTitle: z.string(),
```
anncatton, line 37: is the `title` like `Dr.`, `Ms.` etc? if so wondering if we could have some set values for this field

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -27,6 +27,26 @@ import { Name } from './Name.js';
 import { OhipNumber } from './OhipNumber.js';
 import { NanoId } from './NanoId.js';
 
+export const hasRequiredGuardianInformation = (
```
anncatton, line 30: 👍

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ConsentGroup, ClinicianInvite } from '../../src/entities/index.js';
+
+describe('ClinicianInvite', () => {
+	it('Must define conditionally required fields on condition', () => {
+		expect(
+			ClinicianInvite.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
+				inviteSentDate: new Date(),
+				clinicianFirstName: 'Homer',
+				clinicianLastName: 'Simson',
```
anncatton, line 12: Obviously not important but i can't unsee a typo...esp if it relates to a simpsons ref 😂 

(Please ignore this dumb comment and don't change lol)

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { expect } from 'chai';
+
+import { ConsentGroup, ClinicianInvite } from '../../src/entities/index.js';
+
+describe('ClinicianInvite', () => {
+	it('Must define conditionally required fields on condition', () => {
+		expect(
+			ClinicianInvite.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
```
anncatton, line 9: can remove the `-` in these test ids since we know we'll be disallowing them in our nanoid setup

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+				consentGroup: ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR,
+				guardianName: 'Marge Simpson',
+				guardianPhoneNumber: '1234567890',
+				guardianEmailAddress: 'marge.simpson@example.com',
+				guardianRelationship: 'Wife',
+			}).success,
+		).true;
+		expect(
+			ClinicianInvite.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
```
anncatton, line 28: remove `-`

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+				participantEmailAddress: 'bart.simpson@example.com',
+				participantPhoneNumber: '6471234567',
+				consentGroup: ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR,
+				guardianName: 'Marge Simpson',
+				guardianRelationship: 'Wife', // missing guardianEmailAddress and guardianPhoneNumber
+			}).success,
+		).false;
+		expect(
+			ClinicianInvite.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
```
anncatton, line 45: remove `-`

anncatton (approved):
Looks awesome! I just added some comments/thoughts/smol quibbles but nothing dire so i'm approving

packages/types/tests/entities/ClinicianInvite.spec.ts
alyssadsouza: 😂 well now I have to change it

packages/types/src/entities/ClinicianInvite.ts
alyssadsouza: yeah that sounds right, I don't think there would be any issue with null = unknown != false in this case because they either accepted or they haven't. I'll update the prisma schema too then.

packages/types/src/entities/ClinicianInvite.ts
alyssadsouza: checked with @b-f-chan, should just be a free text field

anncatton (approved):
🚀 🚀 🚀

packages/types/tests/entities/ClinicianInvite.spec.ts
anncatton: lol

Issue #86, opened 2023-09-12
[Create ZOD Schema] Participant Response


b-f-chan:
**EXIT CRITERIA:**

- [ ] ZOD validation schema exists with all required validation rules for the fields in the Participant Response table
- [ ] Exported Typescript type for Participant Response generated by Zod.infer

@mistryrn please cleanup the technical language here

**NOTE: For validation, please be sure to cover these aspects**

- [ ] Conditional requirements/mandatoriness of the field
- [ ] Data type
- [ ] Permissible values

b-f-chan:
Merged, closing

Issue #107, opened 2023-09-26
Implement ID Generator for DAS


b-f-chan:
Implement service that auto generates IDs for the various DASes (Data Access Services) in the OHCRN platform

We have chosen to use the Nano ID library: https://www.npmjs.com/package/nanoid

Example of how to implement: https://github.com/prisma/prisma/discussions/6930#discussioncomment-6252888

**EXIT CRITERIA:**

- [ ] Each DAS is using Nano ID to generate unique IDs as required in their data model
- [ ] Update the seed data
- [ ] Update the default ID alphabet to remove hyphens and underscores
- [ ] Ensure when passing in an ID from another database as a prop, don't let the ID generator overwrite it, use the one that was passed in

@mistryrn - Please add technical details to the required exit criteria if needed

b-f-chan:
Note: Latest version of this library doesn't play well with Typescript --> Will continue investigating using latest version without obstacles, else will use older version (but check that it is still feature complete)

b-f-chan:
ESM issue is related to this ticket --> Need to make final decision on the ESM approach

b-f-chan:
Approved, but minor changes pending --> @anncatton to close when ready

b-f-chan:
Merged and closed

PR #149, opened 2023-10-12
🦺 #86: add zod schema for participant response

samrichca:
- add zod schema for participant response
- move existing unit tests to separate files


packages/types/src/entities/ParticipantResponse.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import * as z from 'zod';
+
+export const ParticipantResponse = z.object({
+	// TODO: change from .cuid() to nanoid regex once nanoid is implemented
+	id: z.string().cuid(),
+	consentQuestionId: z.string().trim(),
+	// TODO: change from .cuid() to nanoid regex once nanoid is implemented
+	participantId: z.string().cuid(),
+	response: z.string().trim(),
```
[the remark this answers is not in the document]
alyssadsouza: double checked and boolean is correct

packages/types/tests/entities/ParticipantIdentification.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import { expect } from 'chai';
+
+import { ConsentGroup, ParticipantIdentification } from '../../src/entities/index.js';
+
+describe('ParticipantIdentification', () => {
+	it('Must define conditionally required fields on condition', () => {
+		expect(
+			ParticipantIdentification.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
```
anncatton, line 28: I'll be implementing the custom dictionary that enforces this in #107 , but can remove the `-` from the ids in this file since we'll be disallowing them in our nanoid setup and doesn't look like that is part of this test

packages/types/tests/entities/ParticipantResponse.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { expect } from 'chai';
+
+import { ParticipantResponse } from '../../src/entities/index.js';
+
+describe('ParticipantResponse', () => {
+	it('Must have a consent question', () => {
+		expect(
+			ParticipantResponse.safeParse({
+				id: 'CVCFbeKH2Njl1G41-vCQm',
```
anncatton, line 28: Same as above comment, remove `-`

anncatton (commented):
Nice and clean, thanks for splitting this up! 🎉 
Wondering if we should add some `null` and `undefined` value tests in `Name`, `OhipNumber`, `PhoneNumber` and `PostalCode` just to cover all our bases?

packages/types/src/entities/ParticipantResponse.ts
```diff
@@ -0,0 +1,31 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { z } from 'zod';
+
+import { NanoId } from './NanoId.js';
+
+export const ParticipantResponse = z.object({
+	id: NanoId,
+	consentQuestionId: z.string().trim(),
```
alyssadsouza, line 26: this is nit-picky but I wonder if we should make a custom Zod schema for ConsentQuestion ids? depending on if we will use this in other schemas for endpoint requests, etc. it might be a good idea.

alyssadsouza (commented):
Other than my small comment, looks good to me! Also like @anncatton mentioned I know this isn't part of your ticket but while you're already moving the tests to new files, do you think you could also add an undefined/null test for each of the custom Zod schemas we have (Name, OhipNumber, PhoneNumber, PostalCode)? It could also be done in another PR too though.

packages/types/src/entities/ParticipantResponse.ts
anncatton: hmm good question...i feel like it shouldn't be necessary as the zod schema stuff is more for inputted/unknown values, and the consent question ids will be coming from a dictionary that we define.
But to your point about reusability, maybe it doesn't need to be a zod schema, but at least a shared type!

samrichca: can one of you make a PR or ticket for that? I'm not 100% sure what you mean

anncatton: np, i will confer with @mistryrn

packages/types/tests/entities/OhipNumber.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+		expect(OhipNumber.safeParse('1234567890').success).true;
+		expect(OhipNumber.safeParse('1').success).false;
+		expect(OhipNumber.safeParse('12345678901').success).false;
+	});
+	it('Can only contain numbers', () => {
+		expect(OhipNumber.safeParse('123-456-78').success).false;
+		expect(OhipNumber.safeParse('123 456 78').success).false;
+		expect(OhipNumber.safeParse('#123456789').success).false;
+		expect(OhipNumber.safeParse('123456789.').success).false;
+		expect(OhipNumber.safeParse(undefined).success).false;
```
anncatton, line 35: Add a `null` test too!

packages/types/tests/entities/PhoneNumber.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+		expect(PhoneNumber.safeParse('1').success).false;
+		expect(PhoneNumber.safeParse('12345678901').success).false;
+	});
+	it('Can only contain numbers', () => {
+		expect(PhoneNumber.safeParse('123-456-78').success).false;
+		expect(PhoneNumber.safeParse('+123456789').success).false;
+		expect(PhoneNumber.safeParse('123 456 78').success).false;
+		expect(PhoneNumber.safeParse('+1 (234) 5').success).false;
+		expect(PhoneNumber.safeParse('123456789.').success).false;
+		expect(PhoneNumber.safeParse(undefined).success).false;
```
anncatton, line 36: `null` too plz

packages/types/tests/entities/PostalCode.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+import { expect } from 'chai';
+
+import { PostalCode } from '../../src/entities/index.js';
+
+describe('PostalCode', () => {
+	it('Must be 6 characters long', () => {
+		expect(PostalCode.safeParse('T4B0V7').success).true;
+		expect(PostalCode.safeParse('T4B0V7A').success).false;
+		expect(PostalCode.safeParse('T4B0V').success).false;
+		expect(PostalCode.safeParse(undefined).success).false;
```
anncatton, line 29: `null` too plz!

anncatton (commented):
Just request for `null` tests alongside `undefined`

packages/types/tests/entities/ParticipantResponse.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+				participantId: '0v2jwozojfDVQAXIMZJfs',
+				response: true,
+			}).success,
+		).false;
+	});
+	it('Must have a response', () => {
+		expect(
+			ParticipantResponse.safeParse({
+				id: 'qjVNbQwUdWmddU8AyLoJn',
+				consentQuestionId: 'Sunt amet irure officia Lorem ullamco ex?',
```
samrichca, line 47: consentQuestionId may need to come from an enum

packages/types/tests/entities/ParticipantResponse.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+				participantId: '5yW4tMaJMVef7r_bpcUTF',
```
anncatton, line 48: i missed this one in my previous review, can you remove the `_` char?

anncatton (approved):
Looks good to me. if @alyssadsouza is good with the changes merge away 👍

alyssadsouza (approved):
Looks good to me! 🚀

Issue #150, opened 2023-10-13
Create Universal Layout with Footer


b-f-chan:
2 tasks

- [ ] Create universal layout with footer that will be applicable to all UI pages in the Consent Portal
- [ ] Create an OPT-IN layout that uses the universal template + existing header/navigation/page structure

**Expected Outcome(s)**

- [ ] All existing pages look the same, no visual difference
- [ ] There will be 2 layouts, universal vs. opt-in with optional header

samrichca:
revision to this ticket:

register & invite pages do have a navbar, but it's hidden at desktop size. I added the option to hide the navbar at desktop size depending on the page's pathname.

b-f-chan:
@b-f-chan @patrickdsantos - verify on QA

patrickdsantos:
Great work Sam! Footer looks good to me :)

patrickdsantos:
One note however, the mobile view should have a hamburger menu where the "hello" is located. And for tablet view it should show two buttons "Register" and "Login" instead of Hello.


[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2krQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7ecec55f965cf5a59b21f9ccf2e07e480e1b8d9d/Screenshot%202023-11-09%20at%2011.04.10%20AM.png)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN20rQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d6ed3d4e7649317a1c6eade633c9b6f1a0a56c78/Screenshot%202023-11-09%20at%2011.05.03%20AM.png)

b-f-chan:
@samrichca can this be closed?

PR #156, opened 2023-10-16
#150: hide navbar on desktop on some pages

samrichca:
this PR makes the navbar hidden on desktop, depending on pathname


apps/consent-ui/src/components/Header/HeaderWrapper.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+'use client';
+
+import clsx from 'clsx';
+import { ReactNode } from 'react';
+import { usePathname } from 'next/navigation';
+
+import styles from './Header.module.scss';
+
+// TODO i18n - we should have a dictionary of english & french internal paths
```
anncatton, line 28: i don't think i'm clear on what this comment means? we have a json file for the translated routes [here](https://github.com/OHCRN/platform/blob/develop/apps/consent-ui/src/i18n/routes/routesByLocale.json) - and that's what is used by the `LocalizedLink` component to determine translated pathnames. But you might mean something different?

apps/consent-ui/src/components/Header/HeaderWrapper.tsx
```diff
@@ -0,0 +1,46 @@
...
+const PATHNAMES_WITHOUT_DESKTOP_HEADER = ['/en/invite', '/en/register'];
+
+const checkHiddenOnDesktop = (pathname: string) => {
+	// TODO this may become more complex when URL parameters are added
+	return PATHNAMES_WITHOUT_DESKTOP_HEADER.includes(pathname);
```
anncatton, line 33: Looking at your comment, can you modify this function to ignore url parameters? (unless there is a scenario where they would affect header visibility)

samrichca: I confirmed that parameters won't be an issue so I removed the comment

apps/consent-ui/src/components/Header/HeaderWrapper.tsx
samrichca: thanks for letting me know, updated the PR

alyssadsouza (commented):
Looks good to me! I was thinking instead of hiding the header we could just add some custom styles to modify the spacing because a header technically does show up in the pages you've hidden it for. However the `sticky` positioning would complicate that, it's probably better to just add a secondary header component for those pages when we get to the UI for that.

samrichca:
@alyssadsouza the two pages that don't have a header will have a custom layout that goes all the way to the top, so I left out spacing for now 👍

alyssadsouza (dismissed):

apps/consent-ui/src/components/Header/HeaderWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ValidLanguage } from 'src/i18n';
+import routesByLocale from 'src/i18n/routes/routesByLocale.json';
+
+import { getLinkNameByPath } from '../Link/utils';
+
+import styles from './Header.module.scss';
+
+const ROUTES_WITHOUT_DESKTOP_HEADER = [routesByLocale.en.invite, routesByLocale.en.register];
+
+const checkHiddenOnDesktop = (pathname: string, currentLang: ValidLanguage) => {
```
anncatton, line 34: Does this work for the French paths as well? it might be simpler to check via the path keys `invite` and `register`, instead of directly accessing the English paths 🤔

samrichca: it did work for french but you have the right idea using path keys instead. updated + added a comment

anncatton (approved):

PR #162, opened 2023-10-23
🏷️ 159 - dictionary typing

anncatton:
Converts UI dictionaries to Typescript.

**Consent UI**
- all dictionary files in `i18n/locales/` changed from JSON to TS
- English dictionary structure will be the source of truth for expected types; any key or namespace added in `/locales/en` dictionaries should trigger TS errors in `/locales/fr` and vice versa. 
- uses TS `satisfies` to preserve type structure: 
```
const frenchCommonDictionary = { ... } satisfies CommonDictionary; // this type comes from the English common dictionary
```
- 🚨 `getTranslation` no longer required to be async but **should still be used server side only** 🚨  as the entire dictionary for a language is loaded on function call. This should also protect against any discrepancy in current language between server and client. Will need to implement a way to enforce translations in server components only (TODO added).
- 🚨  **namespace and key are now both required args for `getTranslation`** 🚨 
- Expands typing for all dictionaries + related functions. calling`getTranslation` should now provide autocomplete for namespace and dictionary key options in Intellisense
- uses available enums from Types package as dictionary keys; as other enums are added the remaining enum keys should be updated.
- dictionary files are split up on a per-component basis except for enums which are each treated as their own dictionary. Empty dictionaries are provided for routes/components that still need to be implemented.
- Adds a `TODO` for handling missing translations - need to discuss whether this should trigger a visible UI error
- some unrelated formatting that was discovered when running the build script (`RecaptchaCheckbox` and `RegistrationForm`)

**Types package**
- adds `GLOBAL_REGEX_FLAG` const


apps/consent-ui/src/i18n/hooks/getTranslation.ts
```diff
@@ -17,46 +17,32 @@
...
-		let translation = dictionary[key];
-		// TODO: use this for nested keys, not sure if we will use this approach? namespaces may be sufficient
-		// .split('.')
-		// .reduce((obj, key) => obj && obj[key], dictionary);
+const replaceParams = (
+	original: string,
+	replacements?: Record<string, string | number>,
+): string => {
+	return Object.entries(replacements || {}).reduce((acc, [key, value]) => {
+		const tagRegex = new RegExp(`{{[\\s]*${key}[\\s]*}}`, GLOBAL_REGEX_FLAG);
```
mistryrn, line 30: Can we maybe add a comment explaining what's going on here? For the sleep deprived amongst us 😅

apps/consent-ui/src/i18n/hooks/getTranslation.ts
```diff
@@ -17,46 +17,32 @@
...
-				`Could not find "${language}" translation for key "${key}" in namespace "${namespace}"`,
-			);
-			return key;
-		}
-		// for interpolation
-		if (params && Object.entries(params).length) {
-			Object.entries(params).forEach(([key, value]) => {
-				translation = translation!.replace(`{{ ${key} }}`, String(value));
-			});
+// TODO: is there a way to enforce this function for server side use only?
```
mistryrn, line 35: Good question 🤔 Maybe something Ciaran might know or have some ideas about?

apps/consent-ui/src/i18n/locales/en/index.ts
```diff
@@ -0,0 +1,58 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import ancestryDictionary from './ancestry';
+import birthSexDictionary from './birth-sex';
+import commonDictionary from './common';
+import consentsStep1Dictionary from './consents-step-1';
```
mistryrn, line 23: Super nit-picky – how do we feel about going from `consentsStepXDictionary` and `consents-step-X` to `consentStepXDictionary` and `consent-step-X` (making the `consent` singular)?

apps/consent-ui/src/i18n/locales/en/invite.ts
```diff
@@ -0,0 +1,34 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { ConsentGroup } from 'types/entities';
+
+const dictionary = {
+	[ConsentGroup.enum.ADULT_CONSENT]: 'Adult Consent (>18)',
```
mistryrn, line 23: I like the use of the shared types enum here -- is this the approach we'd like to use for the other enum-related ones? If so, should we make a follow-up ticket for that?

packages/types/src/entities/Regex.ts
```diff
@@ -22,3 +22,5 @@ export const nanoId = /^[A-Za-z0-9_-]{21,36}$/;
 export const ohipNumber = /^[0-9]{10}$/;
 export const phoneNumber = /^[0-9]{10}$/;
 export const postalCode = /^[A-Za-z][0-9][A-Za-z][0-9][A-Za-z][0-9]$/;
+
+export const GLOBAL_REGEX_FLAG = 'g';
```
mistryrn, line 26: Ain't nothin but a G thing 😎

mistryrn (commented):
Thanks for doing this, another massive amount of work 👏 

Looks good, just had a couple questions!

apps/consent-ui/src/i18n/hooks/getTranslation.ts
anncatton: shurr...i will double check with jon cuz he actually wrote this function to replace the previous interpolation setup 😛

apps/consent-ui/src/i18n/locales/en/index.ts
anncatton: oh yeah that sounds nicer, will do!

apps/consent-ui/src/i18n/locales/en/invite.ts
anncatton: ya that was my plan - i had just added a comment in the PR description, but good call on making a ticket!

packages/types/src/entities/Regex.ts
anncatton: lololol 🕶️ 🆒

apps/consent-ui/src/i18n/hooks/getTranslation.ts
anncatton: Possibli...i can pester him

apps/consent-ui/src/i18n/locales/en/invite.ts
anncatton: https://github.com/OHCRN/platform/issues/180

apps/consent-ui/src/i18n/hooks/getTranslation.ts
joneubank: replacing template tags that match `{{ key }}` in the dictionary string

the regex will:
1.  replace every tag with that matches the key (thanks to the global flag)
2. ignore whitespace between the `{{`/`}}` and the tag, so `{{key}}` and `{{ key }}` and all other permutations of spaces are matched

packages/types/src/entities/Regex.ts
joneubank: If exporting for reuse I would recommend the naming pattern that puts the constant category before the name, so: `REGEX_FLAG_GLOBAL`

joneubank:
There is a lot of information about how dictionaries work that is captured in this PR description, and we won't know where to reference it.

Is it possible to start making documentation for these kind of decisions in a markdown file, maybe in a `docs` dir? Like create a file `/docs/translations.md` and copy some of hte description you have here (location of dictionary files, english dict is source of truth for shape, etc.) as well as a quick description for:
1. how to add new translations to dictionary
2. how to use translations (with information about server-side only)
3. how replacements work for translation templates

anncatton:
> There is a lot of information about how dictionaries work that is captured in this PR description, and we won't know where to reference it.
> 
> Is it possible to start making documentation for these kind of decisions in a markdown file, maybe in a `docs` dir? Like create a file `/docs/translations.md` and copy some of hte description you have here (location of dictionary files, english dict is source of truth for shape, etc.) as well as a quick description for:
> 
> 1. how to add new translations to dictionary
> 2. how to use translations (with information about server-side only)
> 3. how replacements work for translation templates

Good call eh, will do

apps/consent-ui/src/components/Footer/Links.tsx
```diff
@@ -20,39 +20,40 @@
...
 import clsx from 'clsx';
 
-import { Translation } from 'src/i18n';
+import { TranslationFunction } from 'src/i18n';
+import { FooterDictionary } from 'src/i18n/locales/en/footer.js';
 
 import styles from './Footer.module.scss';
 
-const footerLinks: { translation: string; url: string }[] = [
+const footerLinks: { translationKey: keyof FooterDictionary; url: string }[] = [
```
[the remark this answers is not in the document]
joneubank: Assuming URLs will eventually come from a config file?

apps/consent-ui/src/components/Footer/Right.tsx
```diff
@@ -20,16 +20,20 @@
...
 
-const Right = ({ translate }: { translate: Translation }) => {
+const Right = ({ translate }: { translate: TranslationFunction }) => {
 	return (
 		<div className={styles.right}>
 			<Link href="#">
-				<Image src={ONGovtLogo} alt={translate('on-govt-logo-alt')} className={styles['on-gov']} />
+				<Image
+					src={ONGovtLogo}
+					alt={translate('footer', 'on-govt-logo-alt')}
```
joneubank, line 34: I am pleased we remembered to translate the alt text too.

apps/consent-ui/src/i18n/hooks/getTranslation.ts
```diff
@@ -17,46 +17,32 @@
...
+export const getTranslation: GetTranslation = (language) => {
+	const dictionary = dictionaries[language];
+	return (namespace, key, params) => {
+		// TODO: consider throwing error if translation not a string/undefined
+		// Decide whether to have a UI error handler for this, and whether failure is at full page or component level
+		// warning log and `|| ''` is currently provided as a stopgap
+		if (!(dictionary && namespace && key)) {
+			console.warn(`Missing translation in ${language} dictionary!`);
```
joneubank, line 43: Good compromise. Log missing cases but don't block UI. Will use this for now.

joneubank (dismissed):

anncatton:
Added a ticket for documentation https://github.com/OHCRN/platform/issues/185

apps/consent-ui/src/i18n/hooks/getTranslation.ts
anncatton: added a jsdoc description for this one

samrichca:
consent step renaming looks good, thanks 🚀

apps/consent-ui/src/i18n/hooks/getTranslation.ts
```diff
@@ -17,46 +17,54 @@
...
+ * Util function that takes the parameters object passed in to the function returned from getTranslation()
+ * and replaces the key in the translated string with the value of that key
+ * Regex will ignore whitespace between the {{/}} and the tag, so {{key}} and {{ key }} and all other permutations of spaces are matched
+ * Uses the global regex flag to ensure each instance of an argument key in a string is replaced
+ *
+ * ```
+ * @param original
+ * @param replacements
+ * @returns string
+ * @example
```
mistryrn, line 36: Thanks for this example! 👏

mistryrn (approved):
Updates look great!

PR #173, opened 2023-10-25
➕ Add zod-openapi tool to types

alyssadsouza:
- added https://github.com/anatine/zod-plugins/tree/main/packages/zod-openapi to `packages/types` and setup the swagger router config in `consent-das` to accept Zod schemas as component schemas in our Swagger JSDocs
  - just implemented in `consent-das` for the `ConsentCategory` and `ConsentGroup` enums in the routers' JSDocs so that we have an example of how to use this in the codebase

**Also note**: it looks like our Zod dependency may be conflicting with the library's own dependencies because every export of a `generateSchema` object must have an explicit type annotation of `SchemaObject`, which is why every import of `generateSchema` is accompanied by `import type { SchemaObject } from 'openapi3-ts/oas31';`, for example
```
import type { SchemaObject } from 'openapi3-ts/oas31';
...
export const ConsentGroupSchema: SchemaObject = generateSchema(ConsentGroup);
```
- it looks like this fix is fine for now so we will move forward with it


apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -146,13 +140,7 @@ router.get('/:consentQuestionId', async (req, res) => {
...
  *                 type: boolean
  *               category:
- *                 type: string
- *                 enum:
- *                   - INFORMED_CONSENT
- *                   - CONSENT_RELEASE_DATA
- *                   - CONSENT_RESEARCH_PARTICIPATION
- *                   - CONSENT_RECONTACT
- *                   - CONSENT_REVIEW_SIGN
+ *                 $ref: '#/components/schemas/ConsentCategory'
```
anncatton, line 143: Nice!

apps/consent-das/src/routers/participants.ts
```diff
@@ -118,13 +118,7 @@ router.get('/:participantId', async (req, res) => {
...
  *                 type: boolean
  *               consentGroup:
- *                 type: string
- *                 enum:
- *                   - ADULT_CONSENT
- *                   - ADULT_CONSENT_SUBSTITUTE_DECISION_MAKER
- *                   - GUARDIAN_CONSENT_OF_MINOR
- *                   - GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT
- *                   - YOUNG_ADULT_CONSENT
+ *                 $ref: '#/components/schemas/ConsentGroup'
```
anncatton, line 121: 👍

anncatton (approved):
This is great! Thanks for adding.

Wondering if it might be useful to add a note/short readme somewhere to describe what this tool is for? (but can be a separate task, i'm actually gonna be adding a `docs` folder to the repo for readmes that are more specific than the general "get started" one)

PR #176, opened 2023-10-25
🦺 Map out /user endpoint in consent-api

alyssadsouza:
Adds a skeleton endpoint for `/user` to retrieve user information in the header
- Added JSDocs to document the endpoint's parameters and responses
- Also updated the `User` Zod schema to describe the response body


apps/consent-api/src/routers/swagger.ts
```diff
@@ -62,6 +63,11 @@ const options = swaggerJsdoc({
 				},
 			],
 		},
+		components: {
+			schemas: {
+				User,
```
samrichca, line 68: hurray, an endpoint in swagger 🎉

samrichca (approved):

packages/types/src/entities/User.ts
```diff
@@ -18,11 +18,21 @@
...
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+
+import { Name } from './Name.js';
+import { UserRole } from './UserRole.js';
 
 export const User = z.object({
-	id: z.string(),
-	name: z.string(),
+	firstName: Name,
```
anncatton, line 28: looks like the `Name` regex allows spaces - if we're splitting up by firstName/ lastName we should probably disallow the space char

One other thought - we should check if our regex allows accented letters

mistryrn: Good catch on the whitespace @anncatton  – I tried to find an official list of what's allowed in Ontario but couldn't find anything definitive.

Looks like as of August 2022 French accent characters are allowed, so we should update the Name regex to include that: https://news.ontario.ca/en/release/1002215/ontario-introduces-french-accents-on-health-cards

Let's leave it as it for now since this will likely require an update to the Data Dictionary. CC @b-f-chan and Linda, looks like our Name regex (for `OHIP First Name`, `OHIP Last Name`, `OHIP middle name`) will need to be updated to allow French accent characters (for example, À, Â, Ç, É, È, Ê, Ë, Ï, Î, Ô, Ù, Û, Ü)

anncatton (commented):
thanks for the detailed followup @mistryrn , this should be good to go for now then
can we add a `TODO` as well on the name regex so that char issue is tracked outside of this pr? Or we can create a placeholder ticket

packages/types/src/entities/Regex.ts
```diff
@@ -19,6 +19,7 @@
 
 import { NANOID_LENGTH, OHIP_NUMBER_LENGTH, PHONE_NUMBER_LENGTH } from './lengthConstraints.js';
 
+// TODO: separate name into two fields with + without whitespace, include French chars in both
```
anncatton, line 22: 👍

anncatton (approved):

PR #191, opened 2023-10-27
🦺 Map out clinician invite form endpoint

alyssadsouza:
- Created the `/invites` router in consent-api
- Added a skeleton endpoint and JSDocs to document the request and response body for `POST /invites` from the Clinician Patient Registration page form submission
  - needed to make `ClinicianInviteBase` schema from `ClinicianInvite` since the form data in the request body (`ClinicianInviteForm`) looks exactly the same with the addition of the `consentToBeContacted` field
  - also updated the `z.date()` types to `z.coerce.date()` so date strings are parsed into `Date` objects where necessary (like in the request body)


packages/types/src/entities/ClinicianInviteForm.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { z } from 'zod';
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+
+import { hasRequiredGuardianInformation } from './ParticipantIdentification.js';
+import { ClinicianInviteBase } from './ClinicianInvite.js';
+
+export const ClinicianInviteForm = ClinicianInviteBase.extend({
+	inviteSentDate: z.coerce.date().default(new Date()), // overwrites base schema's inviteSentDate to set a default only at form submission
```
mistryrn, line 28: Thanks for adding this comment here 🙇

mistryrn (commented):
This looks great!

One thought that just occurred to me for the validation: the actual generation of the Invite ID doesn't occur until after the request is sent (specifically, when the new Clinician Invite record is created in the DASes). So I think the `ClinicianInviteForm` schema may need to somehow mark that `id` as optional in order to pass validation or exclude it when it extends the `ClinicianInviteBase`. I'm not sure what the best way to do that is, open to thoughts/suggestions.

This will likely also impact the registration form for Participant ID

alyssadsouza:
> One thought that just occurred to me for the validation: the actual generation of the Invite ID doesn't occur until after the request is sent (specifically, when the new Clinician Invite record is created in the DASes). So I think the `ClinicianInviteForm` schema may need to somehow mark that `id` as optional in order to pass validation or exclude it when it extends the `ClinicianInviteBase`. I'm not sure what the best way to do that is, open to thoughts/suggestions.
> 
> This will likely also impact the registration form for Participant ID

I've updated `ClinicianInviteForm` to simply `omit` the `id` field, I think this does what we want very cleanly https://zod.dev/?id=pickomit

mistryrn:
Thanks for getting that `omit` change in! One other thing, maybe we should let the DB set the `inviteSentDate` when the record is created, instead of setting at date at the API layer?

CCing @anncatton for her thoughts on this too 👀 🤔

anncatton:
> Thanks for getting that `omit` change in! One other thing, maybe we should let the DB set the `inviteSentDate` when the record is created, instead of setting at date at the API layer?
> 
> CCing @anncatton for her thoughts on this too 👀 🤔

@mistryrn yeah generally i like to leave any created/updated dates to the db, unless there is some specific reason that we'd want to pass it in. But here, this is the date the invite is *created* --> i'm guessing the flow would be, receive the invite request, set in the db, send the invite email, roll back if there's any errors?

mistryrn:
@anncatton yeah, that flow is correct. And since this is the request to create the invite, I think we should be letting the DB set the `inviteSentDate`

mistryrn (approved):
Looks good! Ping me once the merge conflicts are approved and I'll re-approve :)

mistryrn (approved):
🚀

Issue #85, opened 2023-09-12
[Create ZOD Schema] Clinical Profile


b-f-chan:
**EXIT CRITERIA:**

- [ ] ZOD validation schema exists with all required validation rules for the fields in the Clinical Profile table
- [ ] Exported Typescript type for Clinical Profile table generated by Zod.infer

@mistryrn please cleanup the technical language here

**NOTE: For validation, please be sure to cover these aspects**

- [ ] Conditional requirements/mandatoriness of the field
- [ ] Data type
- [ ] Permissible values

b-f-chan:
Tests failing, still need to finish review

b-f-chan:
Merged, closing

PR #192, opened 2023-10-27
🦺 #85: clinical profile zod schema

samrichca:
- fix failing tests due to IDs being too short
- add zod schemas for clinical profile
- add tests


packages/types/src/entities/ClinicalProfile.ts
```diff
@@ -0,0 +1,49 @@
...
+import { NanoId } from './NanoId.js';
+
+export const ClinicalProfile = z
+	.object({
+		ancestry: Ancestry,
+		birthSex: BirthSex,
+		clinicalProfilePrivateKey: NanoId,
+		familyHistoryOfCancer: HistoryOfCancer,
+		gender: Gender,
+		geneticsClinicVisited: GeneticsClinic.optional(),
```
alyssadsouza, line 36: isn't this required? I think the Prisma schema has it as required

alyssadsouza (commented):
Looks good to me!

anncatton (approved):
Great, thanks for the new pr @samrichca 

If @alyssadsouza is good with this i think we can merge 🚀

packages/types/src/entities/ClinicalProfile.ts
anncatton: yup good catch @alyssadsouza , should be required

samrichca: fixed

alyssadsouza (approved):
Looks good! 🚀 🚀 🚀

PR #193, opened 2023-10-29
📝 185 - Add general docs dir, create translation README

anncatton:
**General:**
- Create a `docs` dir for more feature-specific documentation outside the general README.
- Adds a README for the translation implementation, capturing the info in this [PR description](https://github.com/OHCRN/platform/pull/162) as well as some how-to's

**Consent UI:**
- renames `i18n/hooks` -> `i18n/utils`


docs/translations.md
```diff
@@ -0,0 +1,152 @@
+# Translations
+
+This document describes how text content is translated in the Consent UI.
+
+## Implementation
+
+### Location
+
+All translation-related code is located in the [`/apps/consent-ui/` folder](../apps/consent-ui/src/i18n/), including:
```
mistryrn, line 9: I would maybe consider making the label for the directory link `/apps/consent-ui/src/i18n/` instead of just `/apps/consent-ui/` to be explicit, since the labels in the list below omit the `/src/i18n/` part

mistryrn (commented):
This looks great! The screenshots are a nice touch. Very clear, 10/10 would read again.

Added one note requesting a smol change to make it explicit that we're talking about the `i18n` directory, but other than that it's good to do :)

docs/translations.md
anncatton: oh whoops good catch, will do

anncatton:
> This looks great! The screenshots are a nice touch. Very clear, 10/10 would read again.
> 
> Added one note requesting a smol change to make it explicit that we're talking about the `i18n` directory, but other than that it's good to do :)

i was really hoping for at least a 12/10 here @mistryrn 😛

mistryrn (approved):

PR #199, opened 2023-10-30
🗃️ Data Model v4 updates - Consent DAS

alyssadsouza:
Changed `consentGroup` field to be not nullable as per Data Model 4.0
  - updated the Prisma schema + seed data
  - updated the `create` service function
  - note that the JSDocs were already listing it as required by default


anncatton (approved):
lgtm

mistryrn (approved):

Issue #197, opened 2023-10-30
PHI DAS: Add PRISMA Schema Changes for Data Model v4.0


b-f-chan:
Need to update PRISMA schemas for the PHI DAS to account for latest changes in Data Model v4.0.  Some missed changes from v3.0 may also be included.

Changes as summarized in ticket #163:

- [x] Remove `genetics_clinic` field
- [x] Add new optional fields to `ClinicalProfile` table:
   - [x] `self_reported_clinician_first_name`
   - [x] `self_reported_clinician_last_name`
   - [x] `self_reported_clinician_title_or_role`
   - [x] `self_reported_genetics_clinic_visited`
   - [x] `self_reported_molecular_clinic_visited`
- [x] update seed data
- [x] update jsdocs, create service function to reflect changes

alyssadsouza:
Changed `self_reported_molecular_clinic_visited` to `self_reported_molecular_lab_visited` as per @b-f-chan's directions

b-f-chan:
Merged, closing

PR #201, opened 2023-10-30
🗃️ Data Model v4 updates - PHI DAS

alyssadsouza:
Removed `geneticsClinic` field and added optional fields as per Data Model 4.0:
```
selfReportedClinicianFirstName
selfReportedClinicianLastName
selfReportedClinicianTitleOrRole
selfReportedGeneticsClinicVisited
selfReportedMolecularLabVisited
```
  - updated the Prisma schema + seed data
  - updated the `create` service function
  - updated JSDocs + `create` endpoint params
  - note that `selfReportedMolecularClinicVisited` was updated to `selfReportedMolecularLabVisited` in the ticket #197


apps/phi-das/prisma/migrations/20231027212640_data_model_v4_update/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+  Warnings:
+
+  - You are about to drop the column `geneticsClinicVisited` on the `ClinicalProfile` table. All the data in the column will be lost.
+
+*/
+-- CreateEnum
+CREATE TYPE "MolecularClinic" AS ENUM ('CHILDRENS_HOSPITAL_OF_EASTERN_ONTARIO_OTTAWA', 'HAMILTON_HEALTH_SCIENCES_HAMILTON', 'HOSPITAL_FOR_SICK_CHILDREN_TORONTO', 'KINGSTON_HEALTH_SCIENCES_CENTRE_KINGSTON', 'LONDON_HEALTH_SCIENCES_CENTRE_LONDON', 'NORTH_YORK_GENERAL_HOSPITAL_TORONTO', 'SINAI_HEALTH_SYSTEM_TORONTO', 'TRILLIUM_HEALTH_PARTNERS_MISSISSAUGA', 'UNIVERSITY_HEALTH_NETWORK_TORONTO');
+
+-- AlterTable
+ALTER TABLE "ClinicalProfile" DROP COLUMN "geneticsClinicVisited",
```
anncatton, line 11: i know at this point we don't have a real db setup so this kind of migration isn't yet an issue...but when we do have existing data this could cause problems, even just in a dev environment. I'm not sure if the generated sql is a result of using the `migrate-dev` command. At any rate we should start paying attention to the exact steps in these migrations, for example this one should be a `RENAME` of the column, not a `DROP` command

anncatton (approved):
I'm approving, but would love for @mistryrn to take a look too before merging!

apps/phi-das/prisma/migrations/20231027212640_data_model_v4_update/migration.sql
alyssadsouza: from what I've seen so far Prisma will prefer a `DROP` and `ADD` to `RENAME` which is probably why @mistryrn pointed out in the wiki the `--create-only` flag so we can modify the SQL before applying it to the db. that said I was thinking that `geneticsClinic` and `selfReportedGeneticsClinic` were two separate fields, not the same one, mainly because one is nullable and the other isn't. but I definitely agree, what you're pointing out makes sense!

mistryrn (dismissed):
Looks good to me!

alyssadsouza:
closed and reopened for a new branch #205 because the migrations got unnecessarily messy

apps/phi-das/prisma/migrations/20231027212640_data_model_v4_update/migration.sql
anncatton: yeah i just default to `--create-only` because prisma is weirdly aggressive on this...but i guess when the migration commands are being generated like this doing a `DROP` is cleaner than making more complicated assumptions about what the user intended.

Issue #139, opened 2023-10-11
Map Out Consent API Endpoints --> Consent Wizard Step 1


b-f-chan:
Map out the Consent API endpoints & requests needed for the Consent Wizard Step 1

[Mockup Link](https://www.figma.com/proto/uft9F45tddCba46xnL4Tfg/OHCRN%2FDev%2FMVP?type=design&node-id=41-13250&t=HldMEc19CwIOzLdo-0&scaling=min-zoom&page-id=2%3A11926)

**Expected Outcomes:**

- [ ] Swagger Docs listing the required endpoints/requests --> Can be single Wiki page with different sub-section for each UI page
    - [ ] List of specific web requests to be made (e.g. POST)
    - [ ] Type of method used
    - [ ] Query parameters
    - [ ] Message body structure
    - [ ] URLs required
    - [ ] What validations are needed on API side when request is received
    - [ ] Final output --> A ZOD object that encapsulates the BODY of the request (actual code sample)
    - [ ] **Document in the Wiki for each endpoint:**
        - [ ] Request parameters
        - [ ] Request body
        - [ ] Response body

b-f-chan:
Blocked by: https://app.zenhub.com/workspaces/ohcrn-63ce4baf33421e00316b1f7b/issues/gh/ohcrn/platform/164

b-f-chan:
Have not made branch, Wiki page so far, still needs the sample object

b-f-chan:
Assigning to @alyssadsouza as agreed to finish PR review cleanup

PR #202, opened 2023-10-30
🦺 #139 - map out Consent Wizard Step 1 API

samrichca:
- add endpoint for consent wizard step 1 (informed consent)


apps/consent-api/src/index.ts
```diff
@@ -53,6 +54,7 @@ const App = (config: AppConfig) => {
 	app.use('/participant-responses', ParticipantResponseRouter);
 	app.use('/consent-completion', ConsentCompletionRouter);
 	app.use('/recaptcha', RecaptchaRouter);
+	app.use('/wizard/steps/informed-consent', InformedConsentRouter);
```
mistryrn, line 63: This works, but we can take advantage of nested routers in Express to make just a single `app.use('/wizard', WizardRouter);` in `index.ts` (after merging in Alyssa's latest changes in `develop`) instead of having one for each step.

See this gist for reference: https://gist.github.com/zcaceres/f38b208a492e4dcd45f487638eff716c

Essentially, what we can do is have:
- [ ] `consent-api/src/index.ts` uses `WizardRouter`:
  - `app.use('/wizard', WizardRouter);`
- [ ] `consent-api/src/routers/wizard.ts` uses `StepsRouter`:
  - `router.use('/steps', StepsRouter);`
- [ ] `consent-api/src/routers/steps.ts` uses `InformedConsentRouter` (and eventually the rest of the steps):
  - `router.use('/informed-consent', InformedConsentRouter);`
  
This would require:
- [x] merge latest `develop`
- [ ] remove this `app.use('/wizard/steps/informed-consent', InformedConsentRouter);` from `index.ts`
- [ ] create new `StepsRouter` at `src/routers/steps.ts`
- [ ] have `StepsRouter` use `InformedConsentRouter`
- [ ] update `WizardRouter` to use `StepsRouter`

mistryrn (commented):
Looks good! Still working on feedback for the InformedConsent data/type, but for now here's some feedback on the routes and how we can nest them

mistryrn:
After discussing with Ann, I realized I missed giving you some important information for this ticket with regards to the `InformedConsent` type (or any of the `key: value` pairs for consent questions). Sorry about that! I think the key-value discussion happened while I was waiting for my Mac to update, so I was following along via Slack on my phone 😅 

Anyway, we actually *do* have Consent Question IDs, but at the moment they only live in the `consent-das` seed data since at the time we hadn't finalized the `types` structure. Since we have a list of the IDs, we should make an enum for them in the types library and update references to `consentQuestionId` to use that enum.

Also, in cases like this, instead of `COMPLETED_INFORMED_CONSENT: z.boolean(),` we can do `[ConsentQuestionIDs.informedConsent]: z.boolean(),` instead.

Here's the list of Consent Question IDs (from [`apps/consent-das/prisma/seed/seed-data.ts`](https://github.com/OHCRN/platform/blob/develop/apps/consent-das/prisma/seed/seed-data.ts)):
```
informed-consent__read-and-understand
release-data__clinical-and-genetic
release-data__de-identified
research-participation__future-research
research-participation__contact-information
recontact__future-research
recontact__secondary-contact
review-sign__signed
```

We'll make a separate ticket to update Prisma to use an enum for the Consent Question IDs, but for this ticket you can add one in Zod and then update your `InformedConsent` type accordingly:

- [ ] 1. Add Consent Question IDs type in Zod:
	- [ ] create `ConsentQuestionId` enum and type inside `types/src/entities/ConsentQuestion`, export these from the file as well
	- [ ] update `ConsentQuestion.id` to use that `ConsentQuestionId` as its type
	- [ ] update tests to use new `ConsentQuestionId` type
- [ ] 2. Update `InformedConsent` Type:
	- [ ] update `InformedConsent` type to use specific ID from consent question IDs enum (`informed-consent__read-and-understand`)
	- [ ] rename `InformedConsent` type to `InformedConsentBase`
	- [ ] create `InformedConsentRequest` and `InformedConsentResponse` types which extend `InformedConsentBase`
	- [ ] update POST to use `InformedConsentRequest` type
	- [ ] update GET to use `InformedConsentResponse` type
	- [ ] for GET request, return mock object that matches `InformedConsentResponse` type
	- [ ] update tests to reflect new types

We'll have a separate ticket for updating Prisma, but these changes above can all be tackled in this PR since the rest of the Consent Wizard steps will want to use the Consent Question IDs. As for the renaming to differentiate the `InformedConsentRequest` and `InformedConsentResponse` -- even though they're the same for now, this just gives us a little more flexibility in case the UI ends up needing something a little different from the GET requests down the line.

packages/types/src/entities/InformedConsent.ts
mistryrn, line 1: Looks good, thanks for defining the `Request` and `Response` types

apps/consent-das/prisma/seed/seed-data.ts
```diff
@@ -22,6 +22,7 @@ const participants: Prisma.ParticipantCreateInput[] = [
 ];
 
 const consentQuestions: Prisma.ConsentQuestionCreateInput[] = [
+	// TODO change id to ConsentQuestionId.QUESTION_ID
```
mistryrn, line 25: Since we already have a ticket in progress for this you can remove this comment

apps/consent-api/src/routers/pdf.ts
mistryrn, line 1: Ah, sorry! For individual routes like this `GET wizard/pdf` and the `GET wizard/progress` one below, we don't need to break them out into their own routers/files as it's a bit overkill here. If you don't mind, can we revert those back into the `wizard.ts` router?

apps/consent-api/src/routers/progress.ts
mistryrn, line 1: Same comment as PDF router, let's move this back into `wizard.ts`

apps/consent-api/src/routers/steps.ts
mistryrn, line 1: Looks good!

mistryrn (changes requested):
Looks good overall, just a few change requests:
- [ ] revert `pdf` and `progress` back into `WizardRouter`
- [ ] remove comment in `consent-das` seed data

apps/consent-api/src/routers/informedConsent.ts
```diff
@@ -0,0 +1,114 @@
...
+ *             properties:
+ *               data:
+ *                 $ref: '#/components/schemas/InformedConsentRequest'
+ *     responses:
+ *       201:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               type: object
```
alyssadsouza, line 52: I think you can skip this and go straight to the reference, because it looks like the response body here is just the `InformedConsentResponse` schema. So basically:
```
 *     responses:
 *       201:
 *         description: OK
 *         content:
 *           application/json:
 *             schema:
 *               $ref: '#/components/schemas/InformedConsentRequest'
```
If you open the Swagger-UI at http://localhost:8080/api-docs you should be able to verify the response looks like:
```
{
  "INFORMED_CONSENT__READ_AND_UNDERSTAND": true
}
```
instead of:
```
{
  "data": {
    "INFORMED_CONSENT__READ_AND_UNDERSTAND": true
  }
}
```

alyssadsouza: Also not sure if this should apply to the request body, depending on how we want the request body to look - do we want it to be wrapped in a `data` key?

apps/consent-api/src/routers/informedConsent.ts
```diff
@@ -0,0 +1,114 @@
...
+ *     responses:
+ *       200:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               type: object
+ *               properties:
+ *                 data:
+ *                   $ref: '#/components/schemas/InformedConsentResponse'
```
alyssadsouza, line 94: Same as comment above

mistryrn (dismissed):
Thanks for getting all those changes in! This looks good and should be a solid blueprint for the rest of the Consent Wizard steps :)

packages/types/src/entities/ConsentQuestion.ts
```diff
@@ -18,13 +18,22 @@
  */
 
 import { z } from 'zod';
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
 
 import { ConsentCategory } from './ConsentCategory.js';
 
+const CONSENT_QUESTION_IDS = ['INFORMED_CONSENT__READ_AND_UNDERSTAND'] as const;
```
alyssadsouza, line 26: could you add the remaining questions IDs as well?
```
INFORMED_CONSENT__READ_AND_UNDERSTAND
RELEASE_DATA__CLINICAL_AND_GENETIC
RELEASE_DATA__DE_IDENTIFIED
RESEARCH_PARTICIPATION__FUTURE_RESEARCH
RESEARCH_PARTICIPATION__CONTACT_INFORMATION
RECONTACT__FUTURE_RESEARCH
RECONTACT__SECONDARY_CONTACT
REVIEW_SIGN__SIGNED
```

alyssadsouza (changes requested):
Everything else looks good to me, just add in the rest of the question IDs since we know them right now!

alyssadsouza (approved):
Thanks for adding all the question IDs!

PR #214, opened 2023-10-31
🗃️ Set default db value for inviteSentDate

alyssadsouza:
Set default value to current timestamp for `inviteSentDate` in `ClinicianInvite` table


mistryrn (approved):
Thanks for getting this change in so quickly!

PR #217, opened 2023-10-31
🗃️ Add ConsentQuestionId enum

alyssadsouza:
- Added Consent Question IDs type in consent-das Prisma schema:
```
INFORMED_CONSENT__READ_AND_UNDERSTAND
RELEASE_DATA__CLINICAL_AND_GENETIC
RELEASE_DATA__DE_IDENTIFIED
RESEARCH_PARTICIPATION__FUTURE_RESEARCH
RESEARCH_PARTICIPATION__CONTACT_INFORMATION
RECONTACT__FUTURE_RESEARCH
RECONTACT__SECONDARY_CONTACT
REVIEW_SIGN__SIGNED
```
- updated ConsentQuestion.id and ParticipantResponse.consentQuestionId to use new consent question ID enum
- updated seed-data to use enum values
- updated consent-das services to use new enum values for consent question ids
- create migrations

**Note**: the routers will need to be updated after the `Zod` enum is added

### DIFF
--- a/apps/consent-das/prisma/migrations/20231031195803_use_consent_question_id_enum/migration.sql
+++ b/apps/consent-das/prisma/migrations/20231031195803_use_consent_question_id_enum/migration.sql
@@ -1,0 +1,29 @@
+/*
+  Warnings:
+
+  - The primary key for the `ConsentQuestion` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - The primary key for the `ParticipantResponse` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - Changed the type of `id` on the `ConsentQuestion` table. No cast exists, the column would be dropped and recreated, which cannot be done if there is data, since the column is required.
+  - Changed the type of `consentQuestionId` on the `ParticipantResponse` table. No cast exists, the column would be dropped and recreated, which cannot be done if there is data, since the column is required.
+
+*/
+-- CreateEnum
+CREATE TYPE "ConsentQuestionId" AS ENUM ('INFORMED_CONSENT__READ_AND_UNDERSTAND', 'RELEASE_DATA__CLINICAL_AND_GENETIC', 'RELEASE_DATA__DE_IDENTIFIED', 'RESEARCH_PARTICIPATION__FUTURE_RESEARCH', 'RESEARCH_PARTICIPATION__CONTACT_INFORMATION', 'RECONTACT__FUTURE_RESEARCH', 'RECONTACT__SECONDARY_CONTACT', 'REVIEW_SIGN__SIGNED');
+
+-- DropForeignKey
+ALTER TABLE "ParticipantResponse" DROP CONSTRAINT "ParticipantResponse_consentQuestionId_fkey";
+
+-- AlterTable
+ALTER TABLE "ConsentQuestion" DROP CONSTRAINT "ConsentQuestion_pkey",
+DROP COLUMN "id",
+ADD COLUMN     "id" "ConsentQuestionId" NOT NULL,
+ADD CONSTRAINT "ConsentQuestion_pkey" PRIMARY KEY ("id");
+
+-- AlterTable
+ALTER TABLE "ParticipantResponse" DROP CONSTRAINT "ParticipantResponse_pkey",
+DROP COLUMN "consentQuestionId",
+ADD COLUMN     "consentQuestionId" "ConsentQuestionId" NOT NULL,
+ADD CONSTRAINT "ParticipantResponse_pkey" PRIMARY KEY ("id", "participantId", "consentQuestionId");
+
+-- AddForeignKey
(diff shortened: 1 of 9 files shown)


apps/consent-das/prisma/migrations/20231031195803_use_consent_question_id_enum/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+*/
+-- CreateEnum
+CREATE TYPE "ConsentQuestionId" AS ENUM ('INFORMED_CONSENT__READ_AND_UNDERSTAND', 'RELEASE_DATA__CLINICAL_AND_GENETIC', 'RELEASE_DATA__DE_IDENTIFIED', 'RESEARCH_PARTICIPATION__FUTURE_RESEARCH', 'RESEARCH_PARTICIPATION__CONTACT_INFORMATION', 'RECONTACT__FUTURE_RESEARCH', 'RECONTACT__SECONDARY_CONTACT', 'REVIEW_SIGN__SIGNED');
+
+-- DropForeignKey
+ALTER TABLE "ParticipantResponse" DROP CONSTRAINT "ParticipantResponse_consentQuestionId_fkey";
+
+-- AlterTable
+ALTER TABLE "ConsentQuestion" DROP CONSTRAINT "ConsentQuestion_pkey",
+DROP COLUMN "id",
```
alyssadsouza, line 18: I tried to adjust this to `ALTER COLUMN "id" TYPE "ConsentQuestionId";` rather than dropping and adding new columns, but Prisma seemed to not like it:

```
db error: ERROR: column "id" cannot be cast automatically to type "ConsentQuestionId"
HINT: You might need to specify "USING id::"ConsentQuestionId"".
```

mistryrn: Thanks for leaving a note about this. Fortunately we're not dealing with real data the the moment but this is good to keep in mind in the future 🙏

mistryrn (commented):
Looks good! Thanks for getting this in so fast, good call on highlighting that the JSDocs schemas in the routers will need to be updated once the Zod schemas are implemented.

Consider this PR approved from my POV, but I'd like to wait for the Zod types to get added in the other PR so we can update those JSDocs before merging this one

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -110,7 +112,7 @@ router.get('/:consentQuestionId', async (req, res) => {
 	const { consentQuestionId } = req.params;
 	// TODO: add validation
 	try {
-		const question = await getConsentQuestion(consentQuestionId);
+		const question = await getConsentQuestion(consentQuestionId as ConsentQuestionId); // TODO: remove 'as ConsentQuestionId' and parse with Zod type
```
anncatton, line 115: thanks for these `TODO`s!

apps/consent-das/src/routers/participantResponses.ts
```diff
@@ -68,11 +70,12 @@ router.get('/:participantId/:consentQuestionId', async (req, res) => {
 	logger.info('GET /participant-responses/:participantId/:consentQuestionId');
 	const { participantId, consentQuestionId } = req.params;
 	const { sort_order } = req.query;
-
+	// TODO: parse with Zod type
```
anncatton, line 73: thanks for calling this out, we want to avoid `as` a rule 🙂

anncatton (approved):
this looks good to me, thanks for all the `TODO`s, esp on those type casts

apps/consent-das/src/routers/consentQuestions.ts
```diff
@@ -110,7 +110,7 @@ router.get('/:consentQuestionId', async (req, res) => {
 	const { consentQuestionId } = req.params;
 	// TODO: add validation
 	try {
-		const question = await getConsentQuestion(consentQuestionId);
+		const question = await getConsentQuestion(ConsentQuestionId.parse(consentQuestionId));
```
mistryrn, line 113: Small nit-pick (blame @anncatton ) – can we split this into two separate lines? One to parse the `ConsentQuestionId`, one to call `getConsentQuestion`?

I think we're doing both the `get()` and `.parse()` in the same line elsewhere, but we can change that later.

apps/consent-das/src/routers/participantResponses.ts
```diff
@@ -68,11 +69,10 @@ router.get('/:participantId/:consentQuestionId', async (req, res) => {
 	logger.info('GET /participant-responses/:participantId/:consentQuestionId');
 	const { participantId, consentQuestionId } = req.params;
 	const { sort_order } = req.query;
-
 	try {
 		const participant_responses = await getParticipantResponses(
 			participantId,
-			consentQuestionId,
+			ConsentQuestionId.parse(consentQuestionId),
```
mistryrn, line 75: Same comment as above re: parsing separately

anncatton (approved):
APPROVÉ

PR #219, opened 2023-11-01
🦺 Add Zod updates for Participant Lifecycle

alyssadsouza:
- added enum for LifecycleState with the following values:
```
REGISTERED
CONSENTED
IN_PROCESSING
PROFILE_COMPLETE
PUBLISHED
DELETION_REQUESTED
PAUSE_REQUESTED
CHANGE_REQUESTED
APPROVED_FOR_DELETION
PAUSED
DELETED
```
- added currentLifecycleState to ParticipantIdentification zod schema, of type LifecycleState (required, will default to REGISTERED on creation in db)
- added previousLifecycleState to ParticipantIdentification zod schema, of type LifecycleState (optional)
- updated unit tests, fixed failing `ClinicianInvite` unit tests by adding `consentToBeContacted` required field (forgot to update the unit tests when moving `consentToBeContacted` to the `ClinicianInviteBase` schema in #191)


packages/types/tests/entities/ParticipantIdentification.spec.ts
```diff
@@ -45,6 +47,7 @@ describe('ParticipantIdentification', () => {
 			ParticipantIdentification.safeParse({
 				id: 'CVCFbeKH2Njl1G41vCQme',
 				ohipNumber: '1234567890',
+				currentLifecycleState: LifecycleState.enum.APPROVED_FOR_DELETION,
```
mistryrn, line 50: It doesn't make much of a difference at the moment, but down the line the LifecycleState could have implications on what other data is/isn't available for a Participant.

For now, let's leave all of these test cases as `REGISTERED`, and when we get the requirements (likely not until we start work on the CRC dashboard next year) we can have test cases with different states and adjust the rest of their data accordingly

mistryrn (changes requested):
All of this looks good! Just one small change request to set all the states in the test data to `REGISTERED` for now :)

mistryrn (approved):
Looks good! Thanks for getting those changes in :)

Issue #179, opened 2023-10-26
Implement Error Handling


anncatton:
Needed for all backend services
- consistent error response types
- logging

Implementation details TBD

b-f-chan:
To be handled in cleanup tickets, closing

Issue #177, opened 2023-10-26
Logging/Debug Configuration for Prisma Clients


anncatton:
Prisma provides some customization options for log levels at the DB level: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging

Discuss + implement what we can add to complement the service Logger setup.

PR #221, opened 2023-11-02
🦺 Map out Consent Wizard - Step 2

alyssadsouza:
Adds skeleton endpoints for GET and POST requests at `/wizard/steps/consent-to-release-data` for consent wizard step 2 page
  - Added JSDocs to document the endpoint's parameters and responses
  - Added the`ConsentReleaseDataRequest` and `ConsentReleaseDataResponse` Zod schemas to describe the request/response body
    - Didn't need any unit tests (no special logic)


packages/types/src/entities/ConsentReleaseData.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const ConsentReleaseDataQuestion = z.enum(CONSENT_RELEASE_DATA_QUESTIONS);
+
+export const ConsentReleaseDataBase = z.object({
+	[ConsentQuestionId.enum.RELEASE_DATA__CLINICAL_AND_GENETIC]: z.boolean(),
+	[ConsentQuestionId.enum.RELEASE_DATA__DE_IDENTIFIED]: z.boolean(),
+	[ConsentReleaseDataQuestion.enum.RELEASE_DATA__FIRST_NAME]: Name,
+	[ConsentReleaseDataQuestion.enum.RELEASE_DATA__LAST_NAME]: Name,
+	[ConsentReleaseDataQuestion.enum.RELEASE_DATA__GENDER_IDENTITY]: Gender,
+	[ConsentReleaseDataQuestion.enum.RELEASE_DATA__OHIP_NUMBER]: OhipNumber.optional(),
```
alyssadsouza, line 62: “OHIP Number” is marked in the UI as required but I'm assuming that just applies to the UI, the request body can have an undefined OHIP number if they check “don’t have an Ontario health card”. The form validation in the UI can check whether the question was responded to.

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,109 @@
...
+ *       403:
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.get('/', async (req, res) => {
+	logger.info(`GET /wizard/steps/consent-to-release-data`);
+	// TODO: implement when auth layer is ready
+	try {
+		logger.info(`Retrieved informed consent`);
```
samrichca, line 99: change logger message to "retrieved consent to release data"

apps/consent-ui/src/i18n/locales/fr/consent-release-data.ts
```diff
@@ -19,6 +19,23 @@
 
 import { ConsentReleaseDataDictionary } from 'src/i18n/locales/en/consent-release-data';
 
-const dictionary = {} satisfies ConsentReleaseDataDictionary;
+const dictionary = {
```
samrichca, line 22: maybe these keys should use the enum you have below (see the locale/en/common.ts file)

alyssadsouza: ah! I made the enum for that purpose and yet forgot to use it, I knew I would forget something

samrichca: based on what @anncatton said about only updating one service - should we update the dictionary when we build the page in consent-ui?

packages/types/src/entities/ConsentReleaseData.ts
mistryrn: Good callout re: `OhipNumber`

We should probably get confirmation from Brandon regarding the "I don't have an Ontario health card" flow. It's been a while since we discussed it but I think the intention may have been to set it to the special `0000000000` case when someone selects that, allowing us to still meet the "required" field.

Do you mind asking Brandon about that in #ohcrn-dev ?

apps/consent-ui/src/i18n/locales/fr/consent-release-data.ts
anncatton: Yes @samrichca that's def the goal! However, with the consent-api mapping work involving the schema types this kind of overlap will occur, esp since we're using enums in the dictionary keys. I think it's fine to leave those changes in here for now, they are clearly called out in the PR description. The dictionary is a somewhat special case as we don't want to miss adding values as we come across them, (and the changes often look "big" because it's just a lot of text) but perhaps going forward it could be useful to have a secondary pr that is just for dictionary modifications. I'm open to suggestions on what the team thinks is best!

alyssadsouza:
> There's also an unrelated failing test in ClinicianInvite -> looks like we forgot to add the consentToBeContacted field when that got added into the schema. That can be a separate fix tho i think - i'm not sure if that file is modified elsewhere atm

@anncatton that was fixed in `develop` in #219, that's my bad for not pulling changes from `develop`.

anncatton:
> > There's also an unrelated failing test in ClinicianInvite -> looks like we forgot to add the consentToBeContacted field when that got added into the schema. That can be a separate fix tho i think - i'm not sure if that file is modified elsewhere atm
> 
> @anncatton that was fixed in `develop` in #219, that's my bad for not pulling changes from `develop`.

aha ok awesome, thanks for verifying @alyssadsouza

packages/types/src/entities/ConsentReleaseData.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Gender } from './Gender.js';
+import { BirthSex } from './BirthSex.js';
+import { Ancestry } from './Ancestry.js';
+import { HistoryOfCancer } from './HistoryOfCancer.js';
+import { PostalCode } from './PostalCode.js';
+import { GeneticsClinic } from './GeneticsClinic.js';
+import { MolecularLab } from './MolecularLab.js';
+import { OhipNumber } from './OhipNumber.js';
+
+const CONSENT_RELEASE_DATA_QUESTIONS = [
```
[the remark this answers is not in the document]
alyssadsouza: that makes sense, I'll rename this to `ConsentReleaseDataFieldName` instead of `ConsentReleaseDataQuestion` unless @mistryrn disagrees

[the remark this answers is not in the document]
mistryrn: Great idea @anncatton @alyssadsouza ! Yeah let's go with `field names` for the form fields, instead of `questions`. The proposed rename sounds fine to me 👍

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,123 @@
...
+ *           schema:
+ *             $ref: '#/components/schemas/ConsentReleaseDataRequest'
+ *     responses:
+ *       201:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/ConsentReleaseDataResponse'
+ *       401:
```
joneubank, line 57: this is out of scope for your ticket but I am writing it here so we think about it...

we need schemas for error responses, the content of the error response message should be declared in the schema and it should be consistent across requests.

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,123 @@
...
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	logger.info(`POST /wizard/steps/consent-to-release-data`);
+	// TODO: implement when auth layer is ready
+	try {
+		logger.info(`Submitted consent to release data`);
+		const data = ConsentReleaseDataRequest.parse(req.body);
```
joneubank, line 69: If this parse fails then an error is thrown, we should be handling this case and communicating to the requester that the request was rejected due to invalid content. This should result in a response with status 400. The response message should ideally include information about what part of the validation failed but translating the zod parsing is a general task that needs focused work so for now just an error 400 would suffice.

We have an opportunity here to write a function to standardize handling POST request body validation, if you want we can talk about building that out quickly.

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,123 @@
...
+ *               $ref: '#/components/schemas/ConsentReleaseDataResponse'
+ *       401:
+ *         description: Unauthorized. Authorization information is missing or invalid.
+ *       403:
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.get('/', async (req, res) => {
+	logger.info(`GET /wizard/steps/consent-to-release-data`);
```
joneubank, line 110: Thanks for thinking about logging. To keep all our request logging consistent and not require maintaining logger statements we should use a middleware function added to the server router that consistently logs all incoming requests and their response status.

There is a common library used for this called [morgan](https://npmjs.org/package/morgan), I'm not a huge fan of it's log formatting or the fact that it waits to log until the response is complete, but this would be a quick fix that serves our needs like 90% if you want to set that up.

In any case, we should remove this log statement.

joneubank (changes requested):
Lots more great work here. I found a couple items that need some review, let me know if any questions. :)

packages/types/src/entities/ConsentReleaseData.ts
```diff
@@ -0,0 +1,86 @@
...
+	'RELEASE_DATA__ANCESTRY',
+	'RELEASE_DATA__HISTORY_OF_CANCER',
+	'RELEASE_DATA__FAMILY_HISTORY_OF_CANCER',
+	'RELEASE_DATA__RESIDENTIAL_POSTAL_CODE',
+	'RELEASE_DATA__SELF_REPORTED_CLINICIAN_TITLE',
+	'RELEASE_DATA__SELF_REPORTED_CLINICIAN_FIRST_NAME',
+	'RELEASE_DATA__SELF_REPORTED_CLINICIAN_LAST_NAME',
+	'RELEASE_DATA__SELF_REPORTED_GENETICS_CLINIC',
+	'RELEASE_DATA__SELF_REPORTED_MOLECULAR_LAB',
+] as const;
```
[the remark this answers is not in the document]
anncatton: My thought for using enums was, let's have them for consistent naming of keys, which I think comes from my experience with the form field keys and req/res types we had in dac-ui -> the type checking of keys there always felt muddled. However I see what you mean here about the unwieldiness of this, and with the zod parse option we will have that check on object keys out of the box.
However, there are some cases where I'd like to keep enums in use - the `ConsentQuestionId` here lines up with the prisma schema [enums](https://github.com/OHCRN/platform/blob/develop/apps/consent-das/prisma/schema.prisma#L52), and I like that for consistency across contexts.

Out of curiosity, is there a way to import zod enums in such a way that the enum name doesn't need to be prepended each time (i'm thinking like static imports in java)?

apps/consent-api/src/routers/consentReleaseData.ts
anncatton: This is one of those "we have a ticket for this!" cases: https://github.com/OHCRN/platform/issues/179

Error handling + custom error types has been on my mind for awhile now, this ticket is not spec'd out yet but I'd be down with getting to this sooner rather than later.

apps/consent-api/src/routers/consentReleaseData.ts
anncatton: Generally these "map out" tasks have been to lay the groundwork for the api endpoints, but not entirely implement them, including full validation and correct error responses. That is I think where the "build <endpoint>" tickets come in. Currently @mistryrn and I are going through the existing "build" tickets to make sure they're spec'd out correctly - I would love to add in standardized post request handling, maybe we can add as part of these build tickets?

apps/consent-api/src/routers/consentReleaseData.ts
anncatton: These kinds of log statements I think are part of the mapping out of the endpoints, and were not necessarily intended to be kept around (and I see as I say that, that if that's the case they shouldn't be added in the first place) tbh we need to discuss what logging would be useful in the api/dases.

packages/types/tests/entities/ConsentReleaseData.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+	ConsentReleaseDataRequest,
+	ConsentReleaseDataResponse,
+	Gender,
+	GeneticsClinic,
+	HistoryOfCancer,
+} from '../../src/entities/index.js';
+import { MolecularLab } from '../../src/entities/MolecularLab.js';
+
+describe('ConsentReleaseDataRequest', () => {
+	it('Must provide valid fields', () => {
```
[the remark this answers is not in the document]
anncatton: Yeah I don't disagree with this. We can prune these tests, but maybe we can use some of this work as reference for the unit testing needed on the api endpoints.

apps/consent-api/src/routers/consentReleaseData.ts
anncatton: I should add, there's a ticket for [logging in prisma](https://github.com/OHCRN/platform/issues/177), but not for the consent-api

packages/types/src/entities/ConsentReleaseData.ts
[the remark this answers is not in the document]
alyssadsouza: > Out of curiosity, is there a way to import zod enums in such a way that the enum name doesn't need to be prepended each time (i'm thinking like static imports in java)?

After some googling I found that we can destructure the enum values:
```ts
const { RELEASE_DATA__FIRST_NAME } = ConsentReleaseDataFieldName.enum;
...
[RELEASE_DATA__FIRST_NAME] = 'First Name';
```

apps/consent-api/src/routers/consentReleaseData.ts
joneubank: I will be satisfied with removing these log statements for now, we don't need to expand the scope of this ticket to request logging.

apps/consent-api/src/routers/consentReleaseData.ts
joneubank: I disagree here for two reasons:
1. part of the describing or "mapping" of the endpoint should be getting the error responses correct
2. code was put in place to validate the inputs but the errors were not handled properly

We should get this right as we add in the validation code.

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,112 @@
...
+ *         description: Unauthorized. Authorization information is missing or invalid.
+ *       403:
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	// TODO: implement when auth layer is ready
+	try {
+		try {
```
anncatton, line 62: i think instead this should be one try/catch block - i'm not sure with this setup if the outer catch would ever be reached 🤔 . we should be able to check what is returned from the `.parse()` call by checking the `error` type in the catch block, and return a 400 if it's a ZodError, and otherwise return a 500. i'm thinking something like:
```
try {
// parse()
// return 201
} catch(error) {
if (error instanceof ZodError) {
// return 400
else {
// return 500
}
```
there'll be much more info in the ZodError object on what exactly failed to parse, but i think for the purposes of this ticket a general check is fine.

anncatton (commented):
Thanks for cleaning up all the tests we didn't need, this is so much more readable. Just one comment about that try/catch - we can also run that past @joneubank to see if that will address his feedback about the error handling.

apps/consent-api/src/routers/consentReleaseData.ts
alyssadsouza: hmm looks like we'll need to move zod outside of types and into the root folder to import that error

packages/types/src/entities/Error.ts
```diff
@@ -0,0 +1,22 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { ZodError } from 'zod';
+
```
anncatton, line 21: maybe just add a comment here about why we're exporting this, and that it's just the initial portion of implementing error handling.

anncatton (approved):
this looks good to me, let's see if jon is also happy with it 😛

packages/types/src/entities/Error.ts
joneubank: Yeah, what is this about? Why wouldn't other places jsut import directly from zod?

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,111 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { Router } from 'express';
+import { ConsentReleaseDataRequest, ZodError } from 'types/entities';
```
joneubank, line 21: Importing zod from our own module feels weird... we should be directly importing from zod, no?

apps/consent-api/src/routers/consentReleaseData.ts
```diff
@@ -0,0 +1,111 @@
...
+		const data = ConsentReleaseDataRequest.parse(req.body);
+		// TODO: make updates and return modified data
+		res.status(201).send(data);
+	} catch (error) {
+		logger.error(error);
+		if (error instanceof ZodError) {
+			res.status(400).send({ message: 'Bad Request' });
+		} else {
+			res.status(500).send({ message: 'Server error' });
+		}
```
joneubank, line 71: Glad to see this. If this pattern gets repeated a lot, as I suspect it will be, theres probably a use for a function for to do request error handling.

joneubank (approved):

packages/types/src/entities/Error.ts
alyssadsouza: copying from one of my previous comments:
>hmm looks like we'll need to move zod outside of types and into the root folder to import that error
instead of doing that, I'll add an Error.ts schema to types which we can use when we add custom schemas for error handling, and just export the zod error type from there for now

after talking with @anncatton we agreed on not moving the Zod import to the root directory for this, since it is just a temporary method for error handling. I'm guessing any Zod types like `ZodError` will be encompassed in some `Error` schema when we actually implement proper error handling, making imports from Zod here just temporary.

alyssadsouza: I'll add a comment there for clarity though

anncatton (approved):

PR #222, opened 2023-11-02
🦺 #203: map out consent wizard step 3

samrichca:
- add router for consent wizard step 3 get & post requests
- add zod schemas & jsdocs
- no unit tests required


packages/types/src/entities/ConsentResearchParticipation.ts
```diff
@@ -0,0 +1,45 @@
...
+export const ConsentResearchParticipationBase = z.object({
+	[ConsentQuestionId.enum.RESEARCH_PARTICIPATION__CONTACT_INFORMATION]: z.boolean(),
+	[ConsentQuestionId.enum.RESEARCH_PARTICIPATION__FUTURE_RESEARCH]: z.boolean(),
+});
+
+export const ConsentResearchParticipationRequest = ConsentResearchParticipationBase;
+export type ConsentResearchParticipationRequest = z.infer<
+	typeof ConsentResearchParticipationRequest
+>;
+export const ResearchParticipationRequestSchema: SchemaObject = generateSchema(
```
alyssadsouza, line 35: definitely nit picking here but should this also be prefixed with `Consent` i.e. `ConsentResearchParticipationRequestSchema`? if not then for consistency I think I should change to match this in #221.

packages/types/tests/entities/ConsentResearchParticipation.ts
alyssadsouza, line 1: missing the `spec.ts` file extension - I ran the tests and was like ... shouldn't there be 21 tests...? 🕵️

packages/types/src/entities/ConsentResearchParticipation.ts
samrichca: not nitpicking at all, i thought i'd caught all of these. good eye!

packages/types/tests/entities/ConsentResearchParticipation.ts
samrichca: 23 now 🌽

alyssadsouza (commented):
Looks good to me!

packages/types/tests/entities/ConsentResearchParticipation.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+	ConsentQuestionId,
+	ConsentResearchParticipationRequest,
+	ConsentResearchParticipationResponse,
+} from '../../src/entities/index.js';
+
+describe('ConsentResearchParticipationRequest', () => {
+	it('Must use boolean values', () => {
+		expect(
+			ConsentResearchParticipationRequest.safeParse({
+				[ConsentQuestionId.enum.RESEARCH_PARTICIPATION__CONTACT_INFORMATION]: true,
```
anncatton, line 32: thanks for the separate tests for req/res

apps/consent-api/src/routers/consentResearchParticipation.ts
```diff
@@ -0,0 +1,111 @@
...
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.get('/', async (req, res) => {
+	logger.info(`GET /wizard/steps/consent-for-research-participation`);
+	// TODO: implement when auth layer is ready
+	try {
+		logger.info(`Retrieved Consent for Research Participation`);
+		res.status(200).send({
```
anncatton, line 101: same comment as above

packages/types/src/entities/ConsentResearchParticipation.ts
```diff
@@ -0,0 +1,45 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { z } from 'zod';
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+
+import { ConsentQuestionId } from './ConsentQuestion.js';
+
+export const ConsentResearchParticipationBase = z.object({
```
anncatton, line 26: i think you mentioned this in the step 2 PR @alyssadsouza , about destructuring the enum - do you wanna use that in these prs for the `ConsentQuestionId` keys?

apps/consent-api/src/routers/consentResearchParticipation.ts
```diff
@@ -0,0 +1,113 @@
...
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	logger.info(`POST /wizard/steps/consent-for-research-participation`);
+	// TODO: implement when auth layer is ready
+	try {
+		try {
+			const data = ConsentResearchParticipationRequest.parse(req.body);
```
anncatton, line 64: same feedback as in step 2, let's get rid of the nested try/catch and do an error type check to differentiate between a parsing error and a general "something went wrong" 500.

anncatton (changes requested):
Mainly commenting on the try/catch in the `POST`, same as in step 2.

packages/types/src/entities/ConsentResearchParticipation.ts
alyssadsouza: yup good idea!

anncatton: 🙏

anncatton (commented):
this looks good to me, but we'll wait on step-2 being reviewed/merged

anncatton (approved):
lgtm 🚀 🎸 🪨

Issue #204, opened 2023-10-30
Map Out Consent API Endpoints --> Consent Wizard Step 4


samrichca:
Map out the Consent API endpoints & requests needed for the Consent Wizard Step 4 Consent to Recontact

[Mockup Link](https://www.figma.com/proto/uft9F45tddCba46xnL4Tfg/OHCRN%2FDev%2FMVP?type=design&node-id=41-13068&t=HldMEc19CwIOzLdo-0&scaling=min-zoom&page-id=2%3A11926)

**Expected Outcomes:**

- [ ] Swagger Docs listing the required endpoints/requests --> Can be single Wiki page with different sub-section for each UI page
    - [ ] List of specific web requests to be made (e.g. POST)
    - [ ] Type of method used
    - [ ] Query parameters
    - [ ] Message body structure
    - [ ] URLs required
    - [ ] What validations are needed on API side when request is received
    - [ ] Final output --> A ZOD object that encapsulates the BODY of the request (actual code sample)
    - [ ] **Document in the Wiki for each endpoint:**
        - [ ] Request parameters
        - [ ] Request body
        - [ ] Response body

b-f-chan:
Assigning to @alyssadsouza as agreed to finish design & PR review

b-f-chan:
Merged, closing

PR #223, opened 2023-11-02
🦺 #204: map out consent wizard step 4 

samrichca:
- add router for consent wizard step 3 get & post requests
- add zod schemas & jsdocs
- add unit tests


packages/types/src/entities/ConsentRecontact.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { z } from 'zod';
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+
+import { ConsentQuestionId } from './ConsentQuestion.js';
+
+export const ConsentRecontactBase = z.object({
+	[ConsentQuestionId.enum.RECONTACT__FUTURE_RESEARCH]: z.boolean(),
+	[ConsentQuestionId.enum.RECONTACT__SECONDARY_CONTACT]: z.boolean(),
```
alyssadsouza, line 28: this will also need to include the secondary contact info fields that we collect if they choose "Yes" to secondary contact:
image

apps/consent-api/src/routers/consentRecontact.ts
```diff
@@ -0,0 +1,111 @@
...
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	logger.info(`POST /wizard/steps/consent-for-recontact`);
+	// TODO: implement when auth layer is ready
+	try {
+		logger.info(`Submitted Consent for Re-Contact`);
+		res.status(201).send({
```
anncatton, line 62: same feedback as in the step-3 pr: let's define these mock responses as separate vars + `TODO` so it's really obvious it's mock data

apps/consent-api/src/routers/consentRecontact.ts
```diff
@@ -0,0 +1,111 @@
...
+ *       500:
+ *         description: Server error
+ */
+router.get('/', async (req, res) => {
+	logger.info(`GET /wizard/steps/consent-for-recontact`);
+	// TODO: implement when auth layer is ready
+	try {
+		logger.info(`Retrieved Consent for Re-Contact`);
+		res.status(200).send({
+			[ConsentQuestionId.enum.RECONTACT__FUTURE_RESEARCH]: true,
```
anncatton, line 102: same as above

packages/types/src/entities/ConsentRecontact.ts
anncatton: following that we'll have to add tests for this too, thanks for pointing this out @alyssadsouza 

This relates back to my comment on step 2 PR https://github.com/OHCRN/platform/pull/221#discussion_r1381829401 -> i think we should refer to these as `fieldNames` or something similar to differentiate them

apps/consent-api/src/routers/consentRecontact.ts
```diff
@@ -0,0 +1,112 @@
...
+ *         description: Unauthorized. Authorization information is missing or invalid.
+ *       403:
+ *         description: Forbidden. Provided Authorization token is valid but has insufficient permissions to make this request.
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	// TODO: implement when auth layer is ready
+	try {
+		try {
```
anncatton, line 62: same comment for this nested try/catch as in steps 2 and 3 🙂

packages/types/tests/entities/ConsentRecontact.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { describe, expect, it } from 'vitest';
+
+import {
+	ConsentQuestionId,
+	ConsentRecontactBase,
+} from '../../src/entities/index.js';
+
+describe('ConsentRecontactBase', () => {
+	it('Must define secondary contact info if user consented to secondary contact', () => {
```
anncatton, line 28: Following Jon's feedback in [step 2](https://github.com/OHCRN/platform/pull/221#discussion_r1382640542), can we add test descriptions for each scenario that is being tested here? i.e.:

`it('Parses correctly when secondary consent is indicated and required fields are provided')`
`it('Fails to parse when secondary consent is indicated but secondaryContactPhoneNumber is not provided')` etc etc

so that the intention of each test is very clear.

anncatton (changes requested):
Just some smol changes, as in the other step PRs 🙂

anncatton (approved):
Thanks for updating the tests + that `TODO`! 🚀

PR #225, opened 2023-11-03
✏️ Fix Swagger JSDocs typos

alyssadsouza:
- fixed some typos in a couple of the Swagger JSDocs `tag` fields which led to some of the router endpoints being grouped in two spots in the Swagger UI (like some under Participant and others under Participants)

### DIFF
--- a/apps/phi-das/src/routers/clinicalProfiles.ts
+++ b/apps/phi-das/src/routers/clinicalProfiles.ts
@@ -100,7 +100,7 @@ router.get('/:clinicalProfilePrivateKey', async (req, res) => {
  * /clinical-profiles/:
  *   post:
  *     tags:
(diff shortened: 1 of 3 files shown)


apps/phi-das/src/routers/clinicalProfiles.ts
```diff
@@ -100,7 +100,7 @@ router.get('/:clinicalProfilePrivateKey', async (req, res) => {
  * /clinical-profiles/:
  *   post:
  *     tags:
- *       - Clinical Profile
+ *       - Clinical Profiles
```
alyssadsouza, line 103: went with plural since the other services uses plural too

joneubank (approved):

PR #226, opened 2023-11-03
🚑️ Omit inviteSentDate in Clinician Invite services

alyssadsouza:
- leftover instances of `inviteSentDate` that I didn't catch in #214

### DIFF
--- a/apps/consent-das/src/routers/clinicianInvites.ts
+++ b/apps/consent-das/src/routers/clinicianInvites.ts
@@ -132,10 +132,6 @@ router.get('/:inviteId', async (req, res) => {
  *               consentToBeContacted:
  *                 type: boolean
  *                 required: true
- *               inviteSentDate:
- *                 type: string
- *                 format: date
- *                 required: true
  *               inviteAcceptedDate:
  *                 type: string
  *                 format: date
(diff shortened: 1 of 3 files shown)


apps/consent-das/src/routers/clinicianInvites.ts
```diff
@@ -159,13 +155,11 @@ router.post('/', async (req, res) => {
...
 		consentGroup,
 		consentToBeContacted,
-		inviteSentDate,
 		inviteAcceptedDate,
 		inviteAccepted,
 	} = req.body;
 	// TODO: add validation
 	try {
-		const parsedInviteSentDate = new Date(inviteSentDate);
 		const parsedInviteAcceptedDate = inviteAcceptedDate ? new Date(inviteAcceptedDate) : undefined;
```
anncatton, line 163: Just thinking here now, why is `inviteAccepted` passed in as an external arg? Shouldn't the db handle this too? i know right now schema doesn't have that setup, but maybe it should...

Maybe i'm forgetting a step in the invite flow here @mistryrn 🤔

anncatton (approved):
Approving this as my question doesn't need to be addressed in the work here - thanks for the fix!

One other SUPER-picky thing (sorry): i noticed your branch name says `hotfix`, but hotfixes are generally reserved for critical bugs in production, outside a release cycle. this is just a bug fix 🙂

apps/consent-das/src/routers/clinicianInvites.ts
mistryrn: When we were setting up the DAS routers we just said we'd pass everything in to keep it simple, since we hadn't figured out the end-to-end flow yet. When we work on the data-mapper tickets we'll be able to update the DAS routers accordingly.

PR #231, opened 2023-11-06
💄 Refactor landing page card

alyssadsouza:
Some UI refactoring/cleanup before adding the consent card to the Participant Dashboard:
- Moved the `Card` component out of `Home` to `components` directory so it can be reused as a wrapper component
    - Updated the styles to be more generic
    - Added `dropShadow` and `layout` props
- Kept `CardLink` component in `Card` directory, simplified the component by removing the `Icon` prop and adding the icons as children
- Added `LandingPageCard` component in `Home/` specifically for the landing page (moved all the previous styles that weren't reusable to `LandingPageCard.module.scss`)
- Replaced the landing page background image with the image asset source file from the mockup (better quality)
- Added a new `imageOverlay` background gradient to the tailwind config and updated the background image styling in the landing page to use it with the `::before` and `::after` css properties instead of the `.gradientOverlay` div
  - This makes it more similar to the way the background image is masked/filtered in the mockups

### Before
Screenshot 2023-11-06 at 12 20 54 PM

### After
Screenshot 2023-11-06 at 12 19 57 PM


apps/consent-ui/src/components/Home/index.tsx
```diff
@@ -21,23 +21,23 @@ import Image from 'next/image';
 
 import { ValidLanguage, getTranslation } from 'src/i18n';
 import LinkButton from 'src/components/Button/LinkButton';
-import BackgroundImage from 'src/public/background.png';
+import BackgroundImage from 'src/public/landing-page.png';
```
samrichca, line 24: i just noticed this image is 18.7mb 🤯 

i'm gonna look into image optimization. that's huge.

samrichca: i want to do a bit more investigation on image optimization. for now my advice is to stick with smaller images.

alyssadsouza: using a smaller file compressed thanks to @patrickdsantos 💯

samrichca (dismissed):
lgtm!

samrichca (approved):
looks great! :rocket:

PR #232, opened 2023-11-06
💄📱 Add consent card to dashboard

alyssadsouza:
- Added `DashboardComponent` for the `/dashboard` page with the consent card (3 states); added responsive styling
- Refactored the `LocalizedLink` component to use a `LinkButton` instead of `Link` if `role="button"`
  - Also moved all of the common `ButtonProps` like `variant`, `color`, etc into `src/components/Button/types.ts` to be reused across `Button`, `LinkButton`, and `LocalizedLink`

**Disabled**
Screenshot 2023-11-06 at 12 25 51 PM

**Incomplete**
Screenshot 2023-11-06 at 12 26 13 PM

**Complete**
Screenshot 2023-11-06 at 12 26 23 PM

**Responsive styling**
https://github.com/OHCRN/platform/assets/70766739/b6cfa32d-aa0a-43a1-959e-52101ad07f81


apps/consent-ui/src/components/Dashboard/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+			<Card dropShadow="sm" className={clsx(styles.card, styles[consentStatus])}>
+				<div className={styles['consents-img']}>
+					<Image src={ConsentsImage} alt={translate('dashboard', 'review-ohcrn-consents-img')} />
+				</div>
+				<div className={styles.content}>
+					<h2>{translate('dashboard', 'review-ohcrn-consents')}</h2>
+					<p>{translate('dashboard', 'review-consents-description')}</p>
+					<div className={styles['button-container']}>
+						{consentStatus == 'complete' ? (
+							<LinkButton href={''} variant="secondary" color="blue">
```
alyssadsouza, line 46: these will need to be `LocalizedLink`s - still need to figure out a DRY way to apply the same `Button`/`Link` props/styles to the `LocalizedLink` component (like using these components as children in a way that doesn't break the app)

samrichca: good idea, I'll come back to it

our convention in next has been to wrap our styled links/buttons with a next link component, but we weren't doing localization and server/client components before, so may need to reevaluate

apps/consent-ui/src/components/Dashboard/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+const statuses = ['disabled', 'incomplete', 'complete'] as const;
+const consentStatus: (typeof statuses)[number] = statuses[Math.floor(Math.random() * 3)];
+
+const DashboardComponent = async ({ currentLang }: { currentLang: ValidLanguage }) => {
+	const translate = await getTranslation(currentLang);
+	return (
+		<div className={styles.dashboard}>
+			<Card dropShadow="sm" className={clsx(styles.card, styles[consentStatus])}>
+				<div className={styles['consents-img']}>
+					<Image src={ConsentsImage} alt={translate('dashboard', 'review-ohcrn-consents-img')} />
```
samrichca, line 40: not all images need alt text, e.g. if the alt text is redundant with adjacent text. you can use an empty alt attribute `alt=""` instead.

https://accessibleweb.com/question-answer/when-should-i-use-a-null-or-empty-alt-tag/#:~:text=Answer,be%20announced%20to%20the%20user.

alyssadsouza: I guess this applies to background images too then? i.e. the landing page background image

alyssadsouza: removed the alt text for the landing page image too

samrichca: good call, images that are decorative (don't communicate information) also don't need alt tags

apps/consent-ui/src/i18n/locales/fr/dashboard.ts
```diff
@@ -19,6 +19,13 @@
 
 import { DashboardDictionary } from 'src/i18n/locales/en/dashboard';
 
-const dictionary = {} satisfies DashboardDictionary;
+const dictionary = {
+	'review-ohcrn-consents': 'Review OHCRN Consents',
+	'review-ohcrn-consents-img': 'Review OHCRN Consents',
```
samrichca, line 24: unused alt text can be removed from dictionaries

samrichca (approved):
looks good, thanks for making those changes :shipit:  🚀

Issue #186, opened 2023-10-26
Build Sample Clinician Invite Page Using ReactHook Form


b-f-chan:
Before we build real pages, we need to ensure we can make proper use of the ReactHook form by creating a sample page with the following components:

- [x] A text field that is required
- [x] A group of checkbox fields
- [x] ~~A group of radio button fields~~ there's no radios on the invite page
- [x] Another text box field that is conditionally mandatory if you choose the first radio button option
- [x] Select drop-down
- [x] State management example --> E.g. Clicking a button to trigger sending an API request (can be anything)
- [x] Labels and descriptions for accessibility (don't need to add more fields, use these labels against the existing fields above)

If ReactHook form does not work, we will need to evaluate a different tool

**We will use the Clinician Invite Page as the reference example**

**Expected Outcome(s):**

- [x] A sample form with the list above (**NOT** a POC)

samrichca:
Updated scope (discussed on Slack with @b-f-chan @mistryrn)

- [ ] Create an unstyled implementation of the Clinician Invite page
  - [ ] Required & optional text fields
  - [ ] Add a select field (use react-select)
  - [ ] Secondary group of fields appearing if `Consent Group - Option needing guardian` is chosen
  - [ ] Optional: Add a file upload input (if it's not a ton of work)
  - [ ] Add a checkbox
  - [ ] **new** Include ReCaptcha in mock form submission (console log the form submission request)
  - [ ] Add a group of radio inputs, for prototyping purposes

The objective is still to make a prototype of a form, but following a real use case.

samrichca:
additional scope:
- [] update POST /invites API endpoint to handle data as it comes in from the UI
- [] UI & API should use the same zod schema
- [] add more granular error handling
    - [] each field can return its own error(s)
    - [] the response should return errors listed by field
    - [] the message for the error should be the translation key, not translated text
- [] update tests for request and response

samrichca:
Follow up tickets

- https://github.com/OHCRN/platform/issues/315
- https://github.com/OHCRN/platform/issues/300
- https://github.com/OHCRN/platform/issues/313
- https://github.com/OHCRN/platform/issues/289
- https://github.com/OHCRN/platform/issues/291

b-f-chan:
289, 291 - @b-f-chan to close
285, 313, 315 --> @b-f-chan

b-f-chan:
@samrichca can this be closed?

PR #235, opened 2023-11-07
✨➕🚧 #186: add react-hook-form + clinician invite form

samrichca:
### Consent UI

- Add [react-hook-form](https://react-hook-form.com/), [react-select](https://react-select.com/home), and [zod resolver for RHF](https://github.com/react-hook-form/resolvers#zod) for validation.
- Create an unstyled draft of the clinician invite form
- Add dictionaries for clinician invite form & (shared) form labels & errors
- Add generic components for building forms
- Add new README for forms

### Types

Update types & tests for name & phone number regexes: handle empty & whitespace strings

### Special Instructions

Before running these changes, you will need to install react-hook-form, react-select, and @hookform/resolvers:
```
pnpm i
```

## PR Readiness Checklist

- [x] Connected ticket to PR
- [x] New environment variables added to `.env.schema` files, `README.md` - N/A


apps/consent-ui/src/components/Form/FormField/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+		description,
+		error,
+		name,
+		label,
+		required,
+	};
+
+	return (
+		<DefaultFieldSet {...wrapperProps}>
+			<TextInput {...inputProps} />
```
joneubank, line 81: Will this eventually change component used based on the prop `type`?

apps/consent-ui/src/components/Form/DefaultFieldSet/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { ReactNode } from 'react';
+
+const DefaultFieldSet = ({
```
joneubank, line 22: I would remove `Default` from the name. Can specify names later if we have multiple implementations that need differentiation.

packages/types/tests/entities/Name.spec.ts
```diff
@@ -30,5 +30,8 @@ describe('Name', () => {
 		expect(Name.safeParse('Homer_Simpson').success).false;
 		expect(Name.safeParse('-Homer Simpson').success).false;
 		expect(Name.safeParse('Homer Simpson1').success).false;
+		expect(Name.safeParse(undefined).success).false;
+		expect(Name.safeParse(null).success).false;
+		expect(Name.safeParse('').success).false;
```
samrichca, line 35: empty strings pass some zod validation, so we should add them to unit tests when we're also testing undefined & null.

apps/consent-ui/src/i18n/locales/en/index.ts
```diff
@@ -42,6 +45,7 @@ import historyOfCancerDictionary from './history-of-cancer';
 const dictionaries = {
 	ancestry: ancestryDictionary,
 	'birth-sex': birthSexDictionary,
+	'clinician-invite-form': clinicianInviteFormDictionary,
```
justincorrigible, line 48: Note: Not asking for changes.

For my own understanding, what was the reasoning for this project to deviate from the JS-idiomatic camelCase for keys/variables?

apps/consent-ui/src/i18n/locales/en/forms.ts
```diff
@@ -0,0 +1,27 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+const dictionary = {
+	'indicates-required-field': 'indicates required field',
+	'select-placeholder': 'Select an option...',
+} satisfies Record<string, string>;
```
joneubank, line 23: Repeat of previous comment. We want all french dictionaries to satisfy the shape of the english version of the same dictionary. This will let TS enforce that they have the same shape.

apps/consent-ui/src/i18n/locales/fr/form-labels.ts
```diff
@@ -0,0 +1,40 @@
...
+	'first-name': 'Prénom',
+	'guardian-email': 'Addresse e-mail du gardien',
+	'guardian-name': 'Nom du gardien',
+	'guardian-phone': 'Numéro de téléphone du gardien',
+	'guardian-relationship': 'Relation du gardien avec le participant',
+	'last-name': 'Nom de famille',
+	'preferred-name': 'Nom préféré',
+	email: 'Adresse e-mail',
+	phone: 'Numéro de téléphone',
+} satisfies FormLabelsDictionary;
```
joneubank, line 38: Like this! 👍

apps/consent-ui/src/i18n/locales/en/index.ts
samrichca: I'm not sure actually

apps/consent-ui/src/components/Form/FormField/index.tsx
samrichca: the `type` just changes the type attribute in the HTML input

- email inputs use <TextFieldset type="email" /> - they work like a basic text input, but the `type` tells your browser to show an email keyboard (on mobile) and provide email options for autocomplete
- date inputs - <DateFieldset /> with a datepicker library
- checkbox inputs - <CheckboxFieldset />

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+	const {
+		getRecaptchaToken,
+		onRecaptchaChange,
+		recaptchaCheckboxRef,
+		recaptchaError,
+		resetRecaptcha,
+		setRecaptchaError,
+	} = useRecaptcha();
+
+	const handleRecaptchaChangeDemo = () => {
```
mistryrn, line 91: Is this still for demo purposes only, or can we drop the `Demo` suffix from the function name here?

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+		const token = getRecaptchaToken();
+		token && setRecaptchaError('');
+		onRecaptchaChange();
+	};
+
+	// submit form
+	const [successMessageDemo, setSuccessMessageDemo] = useState('');
+	const onSubmit: SubmitHandler<ClinicianInviteFormValidation> = (data, event) => {
+		event?.preventDefault();
+		console.log('SUBMIT DATA', data);
```
mistryrn, line 101: We'll want to clean up the logs before merging, but I assume this is just for testing/dev purposes at the moment so it's fine for now.

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+						<RequiredAsterisk /> {textDict['indicates-required-field']}
+					</p>
+					<TextFieldSet
+						error={errors.participantFirstName?.type && errorsDict['required']}
+						label={labelsDict['first-name'] || ''}
+						name="participantFirstName"
+						required
+					/>
+					<TextFieldSet
+						error={errors.participantFirstName?.type && errorsDict['required']}
```
mistryrn, line 153: Noticed that these are all listed as `errors.participantFirstName?.type`, even for different fields like `participantLastName` – does it make a difference if they're all the same, or should they match up with their respective field `name` values?

apps/consent-ui/src/components/Form/README.md
mistryrn, line 1: Thanks for adding some documentation for this! 😄 Can we move it to the root `docs` directory to keep this kind of technical documentation in the same place?

mistryrn (commented):
Haven't completed my review yet, but leaving a few comments for now before the meeting. Lots of good stuff so far!

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
samrichca: fixed during demo, thanks for pointing it out!

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
samrichca: removed, I'm keeping the success message as demo since it'll be replaced with a redirect to the homepage (added a comment to this effect)

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
samrichca: removed!

packages/types/src/entities/ClinicianInvite/ClinicianInviteForm.ts
```diff
@@ -0,0 +1,71 @@
...
+// this is specifically for validating the Clinician Invite form in consent UI
+// using the zod resolver in react-hook-form.
+// keep it separate from API schemas.
+
+export const ClinicianInviteFormValidation = z
+	.object({
+		clinicianFirstName: Name,
+		clinicianInstitutionalEmailAddress: z.string().email(),
+		clinicianLastName: Name,
+		clinicianTitleOrRole: z.string().trim().nonempty(),
```
alyssadsouza, line 38: I was looking at the docs, is `nonempty` supposed to be for `array` types? https://zod.dev/?id=nonempty I think you'll have to use `min()` here but you probably know more about this since you've been looking into the whole validating empty strings thing

packages/types/src/entities/ParticipantIdentification.ts
```diff
@@ -41,10 +41,19 @@ export const hasRequiredGuardianInformation = (
...
+					guardianName: Name,
+					guardianPhoneNumber: PhoneNumber,
+					guardianRelationship: Name,
+				})
+				.safeParse({
+					guardianEmailAddress,
+					guardianName,
+					guardianPhoneNumber,
+					guardianRelationship,
+				}).success
```
alyssadsouza, line 56: Shouldn't these already be validated from the schema? `.optional()` permits either `undefined` or the specified type, so checking that they're not `undefined` should mean they have the specified type right?

alyssadsouza, line 56: Shouldn't these already be validated from the schema? `.optional()` permits either `undefined` or the specified type, so checking that they're not `undefined` should mean they have the specified type right?

apps/consent-ui/src/components/Form/fieldsets/CheckboxFieldSet/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+				</h4>
+			)}
+			<label htmlFor={name} className="checkbox-fieldset__label">
+				<CheckboxInput required={required} name={name} />
+				<span className="checkbox-fieldset__description">
+					{description}
+					{required && !title && <RequiredAsterisk />}
+				</span>
+			</label>
+			{error && <span style={{ color: 'red' }}>{error}</span>}
```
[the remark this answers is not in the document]
samrichca: I'm gonna do better styling in the form styling ticket, this is bad on purpose ;)

packages/types/src/entities/ClinicianInvite/ClinicianInviteForm.ts
samrichca: you're right, will change to min(1)

apps/consent-ui/src/components/Form/fieldsets/CheckboxFieldSet/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	name,
+	required = false,
+	title,
+}: FormCheckboxFieldSetProps<T>) => {
+	return (
+		<fieldset className={clsx('checkbox-fieldset', className)}>
+			{title && (
+				<h4>
+					{title}
+					{required && <RequiredAsterisk />}
```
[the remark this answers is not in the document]
samrichca: good point 👍

apps/consent-ui/src/i18n/locales/en/index.ts
anncatton: that was me, i also don't remember

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+	const [successMessageDemo, setSuccessMessageDemo] = useState('');
+	// TODO remove and redirect to homepage
+
+	const onSubmit: SubmitHandler<ClinicianInviteFormValidation> = (data, event) => {
+		event?.preventDefault();
+
+		const recaptchaToken = getRecaptchaToken();
+
+		if (recaptchaToken) {
+			axios
```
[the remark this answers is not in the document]
samrichca: any reason for limiting it to 200 originally? we'll need 201 & 204 which IIRC is allowed in axios by default

packages/types/src/entities/ParticipantIdentification.ts
samrichca: after thinking about this more, there's 3 potential outcomes

- undefined (skips non-`optional()` validation)
- defined but doesn't pass validation - short-circuits the validation, `optional()` isn't considered at all.
- defined & does pass validation

so the undefined check is fine and I'll revert to that. empty strings & whitespace are a separate issue that already has a ticket.

apps/consent-ui/src/i18n/locales/en/index.ts
justincorrigible: Not in the scope of this PR, but sounds to me like if there's no practical necessity for deviation from best practices (e.g. a weird third-party library requirement that breaks things otherwise), it'd be best if we return these to consistent JS casing.

Thoughts?

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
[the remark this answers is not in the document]
justincorrigible: This could be abstracted into the response validation options within the Axios client.

That's done at `validateStatus` in the config object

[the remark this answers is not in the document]
samrichca: removed validateStatus as discussed on slack

apps/consent-ui/src/components/Form/RequiredAsterisk/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import clsx from 'clsx';
+
+import styles from './RequiredAsterisk.module.scss';
+
+const RequiredAsterisk = ({ className }: { className?: string; required?: boolean }) => (
```
justincorrigible, line 24: the props `required` is unused and can be removed

apps/consent-ui/src/components/Form/fieldsets/inputs/SelectInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+	required,
+	value,
+}: FormSelectInputProps<T, V>) => (
+	<Select
+		aria-required={required}
+		instanceId={name}
+		name={name}
+		onChange={(val: FormSelectOnChangeArg<V>) =>
+			// the value could be a string or {label, value} object
+			onChange(typeof val === 'string' ? val : val?.value || null)
```
justincorrigible, line 41: 3 thoughts/questions:
1. cognitive burden: this line is quite loaded for this simple validation. It'd be easier to read (and possibly maintain) if abstracted and broken down as a switch.
2. consistency with types elsewhere: I was under the impression we wanted to use `undefined` instead of `null` for empty fields. Personally I'd prefer using nulls (difference between "field has been touched and left empty" vs "untouched field), but still, asking just to double-check for consistency.
3. consistency with types in the same file: as semi-related followup from the last point, in this line the fallback is `null`, while in line 45 the fallback is an empty string. There's likely a reason for this difference, but it isn't clear here. Should this line 41 also default to empty string instead?

apps/consent-ui/src/components/Form/fieldsets/FieldSet/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+	required = false,
+}: FormFieldSetProps<T>) => {
+	return (
+		<fieldset className={clsx('fieldset', className)}>
+			<label htmlFor={name}>
+				{label}
+				{required && <RequiredAsterisk />}
+			</label>
+			{children}
+			{error && <p style={{ color: 'red' }}>{error}</p>}
```
justincorrigible, line 41: Not asking for change, just sharing my own practices that I've found useful:
In layouts that include multiple different children/components, I tend to add blank lines to facilitate visually parsing the layout from source.

```suggestion
			</label>

			{children}

			{error && <p style={{ color: 'red' }}>{error}</p>}
```

apps/consent-ui/src/i18n/locales/en/index.ts
anncatton: ticket: https://github.com/OHCRN/platform/issues/289

anncatton (dismissed):
Approving, but let's have @joneubank / @justincorrigible have final approval if possible

justincorrigible (dismissed):
LGTM 👍 
Thanks for addressing my "consistency" nit picking, @samrichca.

Deferring to @joneubank for final approval

joneubank (changes requested):
- Appreciate you putting so much detail into a docs file to help future devs.
- All the `trims` added to schemas is a good practice.

I feel like we can do better with creating the dictionaries for this form. I would like to see the use or Partial removed at the very least, and maybe a new function that returns an entire dictionary of a given language (rather than rebuilding it by requesting each property individually).

Other than that, I think it would be good to incorporate the type changes I recommended in the other PR.

packages/types/src/entities/Name.ts
```diff
@@ -21,5 +21,5 @@ import { z } from 'zod';
 
 import { name } from './Regex.js';
 
-export const Name = z.string().regex(name);
+export const Name = z.string().trim().regex(name);
```
alyssadsouza, line 24: I was looking at the ticket you added where `Name` and `PhoneNumber` fail validation when the form returns no response as an empty string but 1) the regex doesn't accept an empty string, and 2) optional fields that don't have a value should be `undefined` according to how we treat them in all other schemas #285 

Maybe we could update the regex to allow an empty string, and add a `NameOptional` schema that does a `transform` on empty strings to convert them to `undefined`, like this:

```ts
const name = /^[A-Za-z\s]*$/ // using `*` instead of `+` so it allows 0 occurrences

export const NameOptional = z
	.string()
	.trim()
	.regex(name)
	.transform((x) => x || undefined);
```

Then for the original `Name` schema just add a `min(1)` constraint:
```ts
export const Name = z.string().trim().regex(name).min(1);
```

This could go in a separate PR for that ticket, just leaving it here in case you have any thoughts

apps/consent-ui/src/components/ClinicianInviteForm/types.ts
```diff
@@ -0,0 +1,33 @@
...
+import { FormsDictionary } from 'src/i18n/locales/en/forms';
+import { ConsentGroupDictionary } from 'src/i18n/locales/en/consent-group';
+
+import { FormSelectOption } from '../Form/types';
+
+// TODO ClinicianInviteFormDictionary shouldn't be partial in final version
+// i.e. when styling/page is complete.
+export type ClinicianInviteFormTextDictionary = Partial<
+	ClinicianInviteFormDictionary & FormsDictionary & FormErrorsDictionary
+>;
```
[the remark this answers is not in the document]
samrichca: not finished adding everything to the page yet, but this will be updated by me shortly.

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+	} = useRecaptcha();
+
+	const handleRecaptchaChange = () => {
+		const token = getRecaptchaToken();
+		token && setRecaptchaError('');
+		onRecaptchaChange();
+	};
+
+	// submit form
+	const [successMessageDemo, setSuccessMessageDemo] = useState('');
```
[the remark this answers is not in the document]
samrichca: it can be removed, we can check the network tab to see if the request worked.

packages/types/src/entities/Name.ts
samrichca: thanks alyssa! added to appropriate ticket.

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+				{successMessageDemo && (
+					<Notification level="success" variant="small" title={successMessageDemo} />
+				)}
+			</Form>
+		</FormProvider>
+	);
+};
+
+export default ClinicianInviteFormComponent;
```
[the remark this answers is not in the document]
samrichca: moved my files to components/views & components/common and made an additional ticket

apps/consent-ui/src/app/[lang]/invite/index.tsx
```diff
@@ -18,15 +18,73 @@
...
+import { FormErrorsDictionary } from 'src/i18n/locales/en/form-errors';
 
 const ClinicianRegistration = async ({ currentLang }: { currentLang: ValidLanguage }) => {
 	const translate = getTranslation(currentLang);
+
+	const errorsDict: FormErrorsDictionary = {
+		required: translate('form-errors', 'required'),
+	};
+
+	const labelsDict: Partial<FormLabelsDictionary> = {
```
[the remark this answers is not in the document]
Comment (samrichca) [reply apps/consent-ui/src/app/[lang]/invite/index.tsx:39]: I ran into issues implementing this feedback with the current `kebab-case` in translation dictionaries. I made a ticket for changing kebab to camel, and a follow-up ticket for this form https://github.com/OHCRN/platform/issues/300

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+const ClinicianInviteFormComponent = ({
+	consentGroupOptions,
+	errorsDict,
+	labelsDict,
+	textDict,
+}: {
+	consentGroupOptions: ConsentGroupOption[];
+	errorsDict: FormErrorsDictionary;
+	labelsDict: Partial<FormLabelsDictionary>;
```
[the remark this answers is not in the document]
samrichca: I ran into issues implementing this feedback with the current `kebab-case` in translation dictionaries. I made a ticket for changing kebab to camel, and a follow-up ticket for this form https://github.com/OHCRN/platform/issues/300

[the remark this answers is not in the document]
anncatton: what was the issue you ran into? i'm confused how the key case would affect creating types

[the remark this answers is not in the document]
samrichca: I wanted to deconstruct the translations 

`const { 'kebab-case' } = ...`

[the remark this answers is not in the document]
justincorrigible: Right, I've been thinking about this one, and as a quick stopgap to unblock you without ballooning the scope, you could destructure the ones you can, and separately have things like:
`const camelCase = dictionary['kebab-case'];`

which you could then remove on a cleanup followup PR that also fixes the casing.

apps/consent-ui/src/app/[lang]/invite/index.tsx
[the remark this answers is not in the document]
Comment (anncatton) [reply apps/consent-ui/src/app/[lang]/invite/index.tsx:39]: A followup question to what Jon mentioned - just to make sure I understand the goal with the new ticket: it's to make dictionaries for *each* form as we build them, correct? i.e. a ClinicianInviteFormDict, a ConsentWizardStep2Dict, etc etc? That was my original intention with having dictionaries dedicated to particular components, as opposed to trying to over-generalize with a "FormDictionary" of sorts.

Because even here, the `labelsDict` works with `satisfies FormLabelsDictionary` without the use of `Partial`.

apps/consent-ui/src/components/common/forms/fieldsets/inputs/SelectInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+const SelectInput = <T extends FieldValues, V extends string>({
+	name,
+	onChange,
+	options,
+	placeholder = '',
+	required,
+	value,
+}: FormSelectInputProps<T, V>) => {
+	return (
+		<Select
```
mistryrn, line 36: Not sure if this is a limitation of the `react-select` library, but I can't seem to find where/how it applies `label`s to the select `input`s.

When running the [aXe DevTools](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) I get a Critical Issue relating to the select input created, stating the form element does not have a label:
Screen Shot 2023-12-06 at 4 13 08 PM

Can you look into this and see if there's some way to have it apply an `aria-label` of the `value`, or something like that (or a better solution than that :P)

apps/consent-ui/src/components/views/invite/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,281 @@
...
+import RecaptchaCheckbox from '../../../common/forms/RecaptchaCheckbox';
+
+import { ClinicianInviteFormTextDictionary, ConsentGroupOption } from './types';
+
+const consentGroupsRequiringGuardian: ConsentGroup[] = [
+	ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR,
+	ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT,
+];
+
+const guardianInfoFields: Partial<keyof ClinicianInviteRequest>[] = [
```
anncatton, line 48: Is it possible to reuse the key names from `InviteGuardianFields` here instead?

mistryrn (commented):
Lots of good stuff in here, I really like how modular the form components are! Just a few comments/questions below.

I checked out the branch and built/ran it locally, then did a quick aXe DevTools scan – we might need to take a look at how that `react-select` library works, since it's currently giving us a Critical a11y issue related to input labels (screenshot provided on PR comment).

One of the translations isn't currently showing up for me (`textDict['clinician-information']`), and I wasn't sure about the use of `{ required: true }` in the `rules` prop for the `SelectFieldSet`'s `Controller`.

Other than that, the only major question I had was regarding other errors, for things other than just `required` – I don't see that handled in this PR, but not sure if that was part of the scope or if it has already been discussed.

apps/consent-ui/src/components/ClinicianInviteForm/index.tsx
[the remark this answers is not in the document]
anncatton: seems like in that case it would be better - if possible - to add this deconstruction in after the case change pr is in...i think you already have that work on the go?

apps/consent-ui/src/components/common/forms/fieldsets/inputs/SelectInput.tsx
samrichca: I had to change instanceId to inputId, my b.

apps/consent-ui/src/components/views/invite/ClinicianInviteForm/index.tsx
samrichca: I can't reply directly to your other comment

I'd like to have dictionaries for each form/component but I was thinking we could reuse some of the form labels and have one source of truth for translating them, e.g. we can get "first name/prénom" from a shared dictionary. is that alright or should they be completely separate?

apps/consent-ui/src/app/[lang]/invite/index.tsx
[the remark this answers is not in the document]
Comment (samrichca) [reply apps/consent-ui/src/app/[lang]/invite/index.tsx]: yes, I want to have separate dictionaries for all the forms for type safety, but I'm hoping they can share some of the translations from `FormLabelsDictionary`

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+	// watch consentGroup value & show/hide guardian info fields if participant is a minor.
+	const watchConsentGroup = watch('consentGroup');
+	const [showGuardianFields, setShowGuardianFields] = useState<boolean>(false);
+	useEffect(() => {
+		if (consentGroupsRequiringGuardian.includes(watchConsentGroup)) {
+			// guardian fields are registered on render, in their input components
+			setShowGuardianFields(true);
+		} else {
+			setShowGuardianFields(false);
+			guardianInfoFields.forEach((field: Partial<keyof ClinicianInviteRequest>) => {
```
anncatton, line 124: i think the type on `field` can be removed since `guardianInfoFields` is already typed above

anncatton (commented):
Thanks for all the comments/TODO ticket links @samrichca ! I just added one comment to add one more TODO ticket link, and a minor type change.

I'd like @mistryrn and @justincorrigible to leave a final approval for any changes they requested; i think the latest changes + the followup tickets address the rest of @joneubank 's feedback but i'd like their 👀 there too for a sanity check.

Can we link the followup tickets to the original #186 too?

apps/consent-ui/src/components/common/Form/types.ts
```diff
@@ -0,0 +1,101 @@
...
+export type FormSelectOnChangeArg<V extends string> = SingleValue<string | FormSelectOption<V>>;
+
+// setup types for react-hook-forms API
+type FormFieldName<T extends FieldValues> = Path<T>;
+
+// fieldsets that use the FieldSet component
+
+interface FormFieldSetSharedProps<T extends FieldValues> {
+	className?: string;
+	error?: any; // TODO harmonize error dictionary & RHF errors
```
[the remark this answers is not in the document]
samrichca: added

samrichca:
added follow up tickets to this ticket

mistryrn (commented):
All the recent changes look good! Thanks for highlighting the follow-up tickets where applicable 🙏 

My only outstanding question/comment (apologies if this has already been covered earlier in the PR process) is regarding the optional "Preferred Name" field: have we figured out how to make it actually optional on the form? I'm still getting a "required" message when I try to submit the form without filling out that field:
Screen Shot 2023-12-13 at 1 23 00 PM

samrichca:
@mistryrn very good question, there's another followup ticket for that. https://github.com/OHCRN/platform/issues/285

mistryrn (approved):
Thanks for clarifying/listing out all the follow-up tickets, @samrichca !

Since the follow-up/out-of-scope work is captured in the tickets you listed in this comment (https://github.com/OHCRN/platform/issues/186#issuecomment-1854194859), I'm good to proceed with merging this PR.

Do we have a plan/order/priority for those follow-up tasks?

I think we'll still need @joneubank to give final approval since his prior change request will block you from merging though.

samrichca:
thanks @mistryrn, I've updated that comment with details about priority and status.

if @joneubank isn't available tomorrow morning I'll see if @justincorrigible can merge it

anncatton (approved):

PR #237, opened 2023-11-08
🌽 ✅ chore: add vitest to jenkins

samrichca:
add Test step between Prepare and Build steps


mistryrn (approved):
Thanks for adding this!

PR #241, opened 2023-11-10
🏷️ 180 - enum dictionary typing

anncatton:
Improves typing for enum dictionaries in consent-ui

**Consent-ui**:
- moves `ConsentCategory`, `ConsentGroup` and `HistoryOfCancer` translation keys from the common dictionaries into their own dictionary files
- adds dictionaries for `MolecularLab` enum
- removes direct use of `EnumType.enum.value` as dictionary keys, and instead types every enum dictionary with `satisfies Record<EnumType, string>` instead of `string` for the key to ensure dictionaries match enums in `types` package
- updates component files referencing the modified dictionaries to match the new structure
- **unrelated**: fixes some incorrect formatting in `Header` files

**Types package**:
- adds `MolecularLab` to exports


mistryrn (approved):
Looks good!

Issue #224, opened 2023-11-03
Improve import setup for zod-openapi library 


anncatton:
The current use of `zod-openapi` in the `types` package requires an import of a `SchemaObject` type from the `openapi3-ts/oas31` dependency. This appears to be a known error, showing in Typescript ~4.1.3 and higher. Without the extra import, the build script for `types` will fail with this error:

```
error TS2742: The inferred type of 'UserSchema' cannot be named without a reference to '.pnpm/openapi3-ts@4.1.2/node_modules/openapi3-ts/oas31'. This is likely not portable. A type annotation is necessary.
```

There are a couple possible options for removing the need to import this type for every file that uses the `generateSchema` function:
1) create a file that imports this type anywhere inside the `types` `/src` folder. 
2) create a utils file that imports this type, and exports an explicitly typed version of `generateFunction` to be used by the entity schemas
3) As suggested [here](https://lightrun.com/solutions/microsoft-typescript-the-inferred-type-of-x-cannot-be-named-without-a-reference-to-y-this-is-likely-not-portable-a/), add `preserveSymlinks: true` to the `tsconfig.json` file. This does get rid of the error, but would need to investigate in case this causes unknown side effects.

Issue #178, opened 2023-10-26
Add Validation for DAS layer


anncatton:
Just creating a placeholder ticket - will need to add details. TLDR; we currently have a bunch of `TODO`s for adding validation to requests coming into the DAS endpoints, for example:
```
const { participantId, clinicalInviteId } = req.body;
// TODO: add validation
```

b-f-chan:
Closing, we'll have separate tickets

Issue #72, opened 2023-09-12
Transition from `fetch` to `axios` in Data Mapper


mistryrn:
Switch HTTP requests in Data Mapper from using `fetch` to using `axios`
- https://axios-http.com/

Additionally, take advantage of the `Axios Instance` to avoid repeatedly setting things like Headers in each request:
- https://axios-http.com/docs/instance

Exit Criteria:
- [x] Add the `axios` package to Data Mapper
- [x] Create a shared axios instance for the Data Mapper with default headers such as content-type
- [x] Switch all `fetch` requests in the Data Mapper to use `axios` instead
    - Note: these can be found by searching for the following `TODO` item: `// TODO: use axios instead of fetch`

b-f-chan:
Completed, merged, closing

PR #242, opened 2023-11-10
✨ POC: Add `POST /invites` Data Mapper endpoint

alyssadsouza:
### Description of Changes
**Data Mapper**
- added the `invites.ts` router with the `POST /invites` endpoint, documenting the requests and responses in the JSDocs
- reorganized `services` directory by separating DAS services by files:
  - `services/create.ts` contains all `data-mapper` create services used in the routers 
    - added `createInvite()` to submit clinician invite data to `pi-das` and `consent-das` and return created invite
  - `services/search.ts` contains all `data-mapper` read services used in the routers 
  - `services/das` contains a file for each DAS, which contains all API calls interacting with the specified DAS
    - added `createInviteConsentData()` in `services/consentDas.ts` to POST Consent invite data
    - added `createInvitePiData()` in `services/piDas.ts` to POST PI invite data
```
.
└── services/
    ├── das/
    │   ├── consent.ts  --> createInviteConsentData()
    │   ├── keys.ts
    │   ├── phi.ts
    │   └── pi.ts --------> createInvitePiData()
    ├── create.ts -----------> createInvite()
    └── search.ts
```

**Types package**
- renamed `ClinicianInviteForm` to `ClinicianInviteRequest` more appropriately
- reorganized `ClinicianInvite` schemas into a directory:
  - `ClinicianInvite.ts` contains the base `ClinicianInviteBase` schema that all the rest use, `ClinicianInviteRequest` which is the parameters required to create the invite, and `ClinicianInviteResponse` which is the logical data returned from Consent API for the invite
  - `index.ts` exports all the schemas (excluding `ClinicianInviteBase`) and contains the original `ClinicianInvite` schema
  - `ConsentClinicianInvite.ts` contains the `ConsentClinicianInviteRequest` and `ConsentClinicianInviteResponse` schemas for the data sent to and returned from (having `null` values parsed to `undefined`) `consent-das`
  - `PIClinicianInvite.ts` contains the `PIClinicianInviteRequest` and `PIClinicianInviteResponse` schemas for the data sent to and returned from (having `null` values parsed to `undefined`) `pi-das`


apps/data-mapper/src/services/create/index.ts
```diff
@@ -0,0 +1,94 @@
...
+	});
+
+	return {
+		...participantPiData,
+		ohipNumber: participantOhipNumber,
+		emailVerified: participantConsentData.emailVerified,
+	};
+};
+
+export const createInvite = async ({
```
mistryrn, line 42: Let's add a JSDoc comment for this `createInvite` function explaining the order of operations. I know we haven't done that for the rest of the data-mapper, but with the Clinician Invite flow being the "example" we want to follow, I think that would be a good practice.

Adds context for the business logic i.e. "creates clinician invite in the PI DAS first to get an `inviteId`, then uses the same `inviteId` to create a corresponding entry in the Consent DAS"

apps/data-mapper/src/services/create/index.ts
```diff
@@ -0,0 +1,94 @@
...
+			consentToBeContacted,
+		});
+
+		return dataMapperClinicianInvite.parse({
+			...invitePiData,
+			...inviteConsentData,
+		});
+	} catch (error) {
+		logger.error(error);
+		// TODO: rollback/delete invites already created
```
mistryrn, line 91: Thanks for calling this out!

apps/data-mapper/src/services/create/piDas.ts
```diff
@@ -0,0 +1,82 @@
...
+	participantEmailAddress,
+	participantPhoneNumber,
+	participantPreferredName,
+	guardianName,
+	guardianPhoneNumber,
+	guardianEmailAddress,
+	guardianRelationship,
+}: piClinicianInvite): Promise<ClinicianInvite> => {
+	const { piDasUrl } = getAppConfig();
+	// TODO: use axios instead of fetch
```
mistryrn, line 59: Similar to a previous comment, let's use axios here instead of fetch since this is a new function.

packages/types/src/entities/ClinicianInvite/index.ts
mistryrn, line 1: Really great idea to split up the `ClinicianInvite` types like this! Love it! 👏

mistryrn (commented):
Lots of really great stuff in here, as usual 👏 

Now that @anncatton and I have had the chance to see what some of this will look like based on our initial thoughts for the ticket, we have some new ideas and tasks 😛 Some to be carried out separately, then some for this PR after those are done.

- [x] 1. Can you do the `service` → `services` renaming in a separate PR? Just to keep the (new) changes in this one easier to follow? That PR would **just** be renaming the directory and nothing else (except maybe updating imports if required?)
- [x] 2. Once that's complete, do you mind taking on this ticket to set up Axios in the `data-mapper` https://github.com/OHCRN/platform/issues/72 . Ideally we want to create a shared Axios Instance like the one in `consent-ui` so we don't have to specify things like `headers: { 'Content-Type': 'application/json' } ` on each request.

After that, we can return to this ticket/PR. 

- [ ] 3. Rather than splitting the `services` like `services/create/DAS` and `services/search/DAS`, let's just do `services/DAS`. So you'll have `services/consentDas.ts` containing both GET and POST operations for the DAS, and you'll still have the `create.ts` and `search.ts` for the combined operations like `createInvite()` and `getParticipant()`
  - **Note: when creating these, please leave the existing participant-related functions from the initial data-mapper PoC wherever they were initially – we'll clean them up when working on participant-related tickets**

apps/data-mapper/src/routers/invites.ts
```diff
@@ -0,0 +1,79 @@
...
+ *     description: Submit clinician invite data across PI and Consent DAS
+ *     requestBody:
+ *       required: true
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/ClinicianInviteRequest'
+ *     responses:
+ *       201:
+ *         description: OK
```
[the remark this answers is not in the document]
alyssadsouza: It looks like `ClinicianInvite` encompasses this schema (with the exception of including `inviteSentDate` which @anncatton mentioned is ok to keep for now) so I updated this to use that as a response schema

apps/data-mapper/src/services/create.ts
```diff
@@ -101,3 +111,65 @@ export const createParticipant = async ({
 		emailVerified: participantConsentData.emailVerified,
 	};
 };
+
+/**
+ * Creates clinician invite in the PI DAS first to get an inviteId,
```
anncatton, line 116: this is a good start to adding docs for some of these more complex functions. Can we implement in jsdoc notation so we can take advantage of Intellisense? https://devhints.io/jsdoc

 We don't have all the details for this function yet, but we can get it started.

apps/data-mapper/src/services/create.ts
```diff
@@ -101,3 +111,65 @@ export const createParticipant = async ({
...
+			participantEmailAddress,
+			participantPhoneNumber,
+			participantPreferredName,
+			guardianName,
+			guardianPhoneNumber,
+			guardianEmailAddress,
+			guardianRelationship,
+		});
+		const invitePiData = await createInvitePiData(piData);
+		// TODO: validating consentData should be done before creating invitePiData,
```
anncatton, line 151: i think it's fine to validate the incoming data as the *request data* we're expecting, which would exclude an id. Once we have the response from the pi das operation, we know that the id from that will be a valid nanoid.

For extra protection, the nanoid implementation in the dases does a validation on incoming ids to be sure they match our expected id format -> so the id passed to consent-das will be checked before insertion.

apps/data-mapper/src/services/create.ts
```diff
@@ -101,3 +111,65 @@ export const createParticipant = async ({
...
+			clinicianFirstName,
+			clinicianLastName,
+			clinicianInstitutionalEmailAddress,
+			clinicianTitleOrRole,
+			consentGroup,
+			consentToBeContacted,
+		});
+		const inviteConsentData = await createInviteConsentData(consentData);
+
+		return dataMapperClinicianInvite.parse({
```
anncatton, line 166: i'm wondering if this `parse()` call is overkill here 🤔

anncatton (commented):
Just leaving some general comments @alyssadsouza , there's prob gonna be a few more incoming, i just need to formulate my thoughts a little more 😛 . Thanks for the initial clean up with the dir names and axios!

apps/data-mapper/src/services/create.ts
alyssadsouza: this one is actually necessary because the `dataMapperClinicianInvite` schema converts any null values in optional fields to undefined

apps/data-mapper/src/services/das/consentDas.ts
```diff
@@ -0,0 +1,57 @@
+/*
```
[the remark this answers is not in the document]
alyssadsouza: that sounds good to me!

apps/data-mapper/src/services/create.ts
alyssadsouza: That makes sense! 

As a follow up, to what extent do we need to validate the data before making DAS API calls here?  Now when I look at it, [`piClinicianInvite`](https://github.com/OHCRN/platform/pull/242/files#diff-cbb2736d7974101dd22d8bc185c080bae58fc38dabf5210ce9059249ec3fe1b5R25) and [`consentClinicianInvite`](https://github.com/OHCRN/platform/pull/242/files#diff-54fe9a4b1767888445257666f22fb14eeb9a2a7c02ef60fcaa98fbbb17969737R25) are both subsets of the [`ClinicianInviteRequest`](https://github.com/OHCRN/platform/pull/242/files#diff-993d2430517afd55b7c5c7d55ee17c90a4c1c0c44e0bb088a004314ab15664b2R27) schema, which is already used to validate and parse the request body in the `/invites` endpoint before calling the `createInvite` service with the same request body. Since `createInvite` expects its arguments to be of the type `ClinicianInviteRequest`, is it sort of redundant here to validate `consentData` and `piData` individually before calling `createInvitePiData(piData)` and `createInviteConsentData(consentData)`, since `consentData` and `piData` have already been validated together as a whole?

packages/types/src/entities/ClinicianInvite/consentApi.ts
```diff
@@ -19,12 +19,12 @@
 
 import { z } from 'zod';
 import { generateSchema } from '@anatine/zod-openapi';
-import type { SchemaObject } from 'openapi3-ts/oas31';
```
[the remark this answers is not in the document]
anncatton: Hmm so i created a ticket to track this issue https://github.com/OHCRN/platform/issues/224 , but to your point about consistency it might be better *for now* to continue the import as you were doing before. I hesitate to go the other way and start removing them since that would basically be part of the work in the investigation ticket.

apps/data-mapper/src/services/das/consentDas.ts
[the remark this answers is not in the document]
anncatton: @mistryrn would you be happy with that? considering our history with naming things 😂

apps/data-mapper/src/services/create.ts
anncatton: i am wondering about making a note of this somewhere...not sure if it would belong in types or in the docs or something. Since this is due to pulling `null`s out of prisma/db i think?

apps/data-mapper/src/services/create.ts
anncatton: that sounds legit to me - if you remove the secondary parsing, the args passed to `createInviteConsentData` and `createInvitePiData` should still be good if everything is typed correctly.

apps/data-mapper/src/services/das/consentDas.ts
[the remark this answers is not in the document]
mistryrn: @anncatton @alyssadsouza this sounds good to me! Thanks both :)

packages/types/src/entities/ClinicianInvite/consentApi.ts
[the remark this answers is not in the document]
alyssadsouza: and for the new schemas I should keep adding them in?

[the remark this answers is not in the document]
anncatton: yeah let's keep with that. when we figure out how we're going to handle it, we can clean everything up.

apps/data-mapper/src/routers/invites.ts
```diff
@@ -0,0 +1,78 @@
...
+ *         description: Bad Request
+ *       500:
+ *         description: Server error
+ */
+router.post('/', async (req, res) => {
+	logger.info(`POST /invites`); // TODO: remove
+	try {
+		const data = ClinicianInviteRequest.parse(req.body);
+		const invite = await createInvite(data);
+		const validatedInvite = ClinicianInviteResponse.parse(invite);
```
alyssadsouza, line 66: I'm not sure if this is too excessive, but because `createInvite` returns data parsed with the `TransformClinicianInvite` schema, and the response type for this endpoint is `ClinicianInviteResponse` (which should be the effective output of `TransformClinicianInvite`) I figured it's safer to parse it before returning since they are two separate schemas

apps/data-mapper/src/services/das/consent.ts
```diff
@@ -0,0 +1,56 @@
...
+	clinicianLastName,
+	clinicianInstitutionalEmailAddress,
+	clinicianTitleOrRole,
+	consentGroup,
+	consentToBeContacted,
+}: ConsentClinicianInviteRequest): Promise<ConsentClinicianInviteResponse> => {
+	const { consentDasUrl } = getAppConfig();
+	try {
+		const result = await axiosClient.post(urlJoin(consentDasUrl, 'clinician-invites'), {
+			clinicianInviteId: id,
```
alyssadsouza, line 41: I caught a bug here while testing because in the `consent-das` router the expected request body lists the field as `clinicianInviteId` versus what I had originally been passing `id` -- but it didn't break in `consent-das`, just autogenerated an ID instead of the one being passed here. Just wanted to point it out as a reminder that we still need to add validation in the DAS.

alyssadsouza: linking the ticket here https://github.com/OHCRN/platform/issues/178

apps/data-mapper/src/services/das/pi.ts
```diff
@@ -0,0 +1,57 @@
...
+			participantLastName,
+			participantEmailAddress,
+			participantPhoneNumber,
+			participantPreferredName,
+			guardianName,
+			guardianPhoneNumber,
+			guardianEmailAddress,
+			guardianRelationship,
+		});
+		// converts all nulls to undefined
```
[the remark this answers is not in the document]
alyssadsouza: yeah good idea, should I add a `schemas.md` page?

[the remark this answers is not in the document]
alyssadsouza: added a `schemas.md` doc, I can rename it if you can think of a better name 😛

mistryrn (commented):
Thanks so much for this one – lots of great work and tons of changes/feedback, plus that new doc will be **very** useful down the line 🙌 Consider this approved from my end, but I'm gonna let @anncatton have the final say here ✅

anncatton (approved):
this is so much good work, thank you @alyssadsouza ! 🪨 🚀

PR #247, opened 2023-11-13
➕ Add axios to data-mapper

alyssadsouza:
**Data Mapper**
- Added `axios` and created an `axiosClient` instance to be used across all API calls sharing common headers
- Updated all `fetch` requests (in `create.ts` and `search.ts`) to use the `axiosClient` and `urlJoin`

Before running:
```
pnpm i
```


mistryrn (approved):
🚀

PR #249, opened 2023-11-14
💄📱 Add Modal component

alyssadsouza:
**Consent UI**
Added a responsive Modal component using React Context to enable showing and hiding the modal from any component within the `PageLayout`
  - `ModalProvider` provides the settings for the modal's visibility and config, and renders the actual modal over all `PageLayout` content when `isOpen` is true
  - `useModal()` lets any component in `PageLayout` access the setters for the modal visibility and config
  - When open, the modal closes when clicking outside the modal card, or on pressing the escape key
    - Added the [`react-detect-click-outside`](https://github.com/zhaluza/react-detect-click-outside) library to handle this

**Before running app:**
```
pnpm i
```

**Preview**
Screenshot 2023-11-14 at 2 31 16 PM

**Responsive Styling**
https://github.com/OHCRN/platform/assets/70766739/01c7dfdd-481d-42c9-8fb9-0e588000e55e


apps/consent-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+		[setIsOpen, setConfig],
+	);
+
+	return (
+		<ModalContext.Provider value={value}>
+			{isOpen && (
+				<div className={styles.modal}>
+					<Card className={styles.card} dropShadow="none">
+						{title && <h3>{title}</h3>}
+						{message && <p>{message}</p>}
```
samrichca, line 63: I would change message -> body & use `<p>` only if the body is a string/not a component

apps/consent-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+						{(actionButtonText || cancelButtonText) && (
+							<div className={styles.buttons}>
+								{onCancelClick && (
```
samrichca, line 66: check for cancel click & cancel text here

apps/consent-ui/src/components/Modal/types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type ModalConfig = {
+	title?: React.ReactNode;
+	message?: React.ReactNode;
+	actionButtonText?: React.ReactNode;
+	actionDisabled?: boolean;
+	cancelButtonText?: React.ReactNode;
+	cancelDisabled?: boolean;
+	onActionClick?: React.ComponentProps<typeof Button>['onClick'];
+	actionLink?: React.ComponentProps<typeof LinkButton>['href'];
```
samrichca, line 31: good idea!

apps/consent-ui/src/components/Card/index.tsx
```diff
@@ -17,11 +17,13 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+'use client';
```
samrichca, line 20: why did this became a client component?

alyssadsouza: oh hmmm, when I was working on it eslint complained, but I don't get that error anymore... must've resolved it while I was working on it - I'll revert it

apps/consent-ui/src/components/Modal/index.tsx
alyssadsouza: should the body be a string? the props were based off the Modal props in the ARGO UI kit which is why they're `ReactNode`s but I'm not sure now https://github.com/icgc-argo/uikit/blob/develop/src/Modal/index.tsx#L116

samrichca: if the type of body is ReactNode, then a developer could provide a string or number and it'll work, so it's good to handle component vs. no component

string should be OK for title, good catch

alyssadsouza:
TODO: close modal on Escape key pressed --> want to find a good library to handle key presses in a hook

apps/consent-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+	const value = useMemo(
+		() => ({
+			setIsOpen,
+			setConfig,
+		}),
+		[setIsOpen, setConfig],
+	);
+
+	const ref = useDetectClickOutside({ onTriggered: () => setIsOpen(false) });
```
samrichca, line 58: I'd suggest adding reusable functions for open & close, with cleanup

openModal = setConfig & setIsOpen(true)

closeModal = reset the config, setIsOpen(false)

when adding reusable components/contexts like this, I clear their contents/state upon removal to avoid any issues with stale data or debugging later.

apps/consent-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+	return (
+		<ModalContext.Provider value={value}>
+			{isOpen && (
+				<div className={styles.modal}>
+					<div ref={ref}>
+						<Card className={styles.card} dropShadow="none">
+							{title && <h3>{title}</h3>}
+							{body}
```
samrichca, line 67: can you add a check here - if body isn't an element wrap it in a P tag

apps/consent-ui/src/components/Card/index.tsx
```diff
@@ -21,19 +21,24 @@ import clsx from 'clsx';
...
 }: {
 	layout?: 'column' | 'row';
 	dropShadow?: 'none' | 'sm' | 'base' | 'lg';
 	className?: string;
 	children: React.ReactNode;
+	onClick?: (e: React.SyntheticEvent<HTMLElement>) => any;
 }) => {
 	return (
-		<div className={clsx(styles.card, styles[layout], styles[`shadow-${dropShadow}`], className)}>
+		<div
```
samrichca, line 38: divs aren't keyboard interactable on their own and shouldn't have onClick attributes but I'm not sure which alternative to suggest... what's the use case for having the card clickable here? I don't see it in this PR 👀

alyssadsouza: forgot to clean that up when switching to the `react-detect-click-outside` library for the close modal behaviour, I'll get rid of it 👍

samrichca (approved):
great work! lgtm!

PR #268, opened 2023-11-20
✨ Introduce express-logger middleware

joneubank:
## Description of Changes

Adds a new package called `express-logger` that provides an express middleware to log all incoming requests using our custom logger. This is a wrapper around the library [express-request-logger](https://www.npmjs.com/package/express-requests-logger).

Produces request logs that look like:
```
2023-11-20T04:58:50.717Z info: [Consent-API.ExpressLogger] 
{
  "method": "GET",
  "path": "/consent-questions/",
  "url": "/?category=CONSENT_RECONTACT",
  "query": {
    "category": "CONSENT_RECONTACT"
  },
  "status": 200,
  "duration": 5
}
```

This can be easily added to any Express app with:

```ts
import express from 'express';
import ExpressLogger from 'express-logger';

import logger from './logger.js';

const app = express();
app.use(ExpressLogger({ logger });
```

The express-logger middleware will log every request as it responds. Most logs will be at `info` level. Authorization errors (HTTP status 401 UNAUTHORIZED and 403 FORBIDDEN) will return a `warn` level log. Any system exception that returns a status of 500 or greater will result in an `error` level log.

The middleware has configuration option to exclude logging on specific URL paths. By default all paths relating to the swagger UI page have been excluded (assuming it is using the path `/api-docs`), and additional paths can be added for each app. For example, it may be good to exclude any path that is planned to be used for health checks:

```ts
app.use(ExpressLogger({ logger, excludeURLs: ['status', 'ping', 'health'] }));
```

#### Logger
The custom logger was refactored to allow for some config variables when created. This lets us control log levels for the router and whether a log file is used. 

The Logger also now has a method called `forModule` so we can append new sub modules to the tag at the start of the logging message - this is used by the express-logger middleware to indicate that the log is coming from the express logger, ie. `[Consent-API.ExpressLogger]`. By default the logger will have no module tag, but this PR updates all apps to name their module when the logger is created. It would be good practice to call `forModule` on the logger after importing it into any service/router/file so that log statements form that source can be clearly identified:

```ts
// This written in a hypothetical file complexService.ts:
import serviceLogger from '../logger.js';

const logger = serviceLogger.forModule('ComplexService');
logger.info('just an example'); // 2023-11-20T04:50:43.733Z info: [Consent-API.ComplexService] just an example
```

#### Types
Added a new exported directory in types for common TS utilities. This has some useful tools such as:
- `RecursivePartial<T>` - To make everything in an object optional, recursively not just on the first level of object properties.
- `Keys<T>` - Create a union of the property names from an object
- `Values<T>` - Create a union of the types of the values of an object. 

## PR Readiness Checklist

- [ ] "Expected Outcome(s)" in ticket have been met
- [x] PR feedback has been addressed
- [x] PR branch is up-to-date with target branch and there are no merge conflicts
- [ ] Connected ticket to PR
- [ ] Ticket number included in PR title
- [x] Labels added to PR


apps/consent-das/src/config.ts
```diff
@@ -1,12 +1,13 @@
 import * as dotenv from 'dotenv';
 
-export type AppConfig = {
-	port: string;
-};
-
-export const getAppConfig = (): AppConfig => {
+export const getAppConfig = () => {
 	dotenv.config();
```
ciaranschutte, line 4: doesn't sound like it's an issue but I remember best practice with this lib at least used to be call it asap.
ie. module level, not in a function
posting in case issue occurs and this comment helps

joneubank: you're right. perhaps we can clean up the config files here a bit in another pr, this is working well enough and unmodified in this change.

apps/phi-das/src/config.ts
```diff
@@ -1,12 +1,13 @@
...
-};
-
-export const getAppConfig = (): AppConfig => {
+export const getAppConfig = () => {
 	dotenv.config();
 	return {
-		port: process.env.PORT || '8083',
+		isProduction: process.env.NODE_ENV === 'production',
+		express: {
+			port: process.env.PORT || '8081',
```
mistryrn, line 8: PHI Das defaults to port `8083`
```suggestion
			port: process.env.PORT || '8083',
```

apps/pi-das/src/config.ts
```diff
@@ -1,12 +1,13 @@
...
-};
-
-export const getAppConfig = (): AppConfig => {
+export const getAppConfig = () => {
 	dotenv.config();
 	return {
-		port: process.env.PORT || '8082',
+		isProduction: process.env.NODE_ENV === 'production',
+		express: {
+			port: process.env.PORT || '8081',
```
mistryrn, line 8:
```suggestion
			port: process.env.PORT || '8082',
```

packages/express-logger/src/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+	return {
+		debug: replacementLogFunction,
+		info: replacementLogFunction,
+		warn: replacementLogFunction,
+		error: replacementLogFunction,
+	};
+};
+
+const swaggerUiPaths = ['api-docs', 'swagger-ui', 'favicon'];
```
mistryrn, line 101: 👏

mistryrn (changes requested):
Thanks a bunch for adding this, it's awesome! Just a few default configs that got mixed up in the copy-pasta across DASes, other than that it looks great! 🎉

apps/consent-api/src/logger.ts
```diff
@@ -17,8 +17,15 @@
...
-import createLogger from 'logger';
+import createLogger, { LogLevels } from 'logger';
 
-const logger = createLogger('Consent-API');
+import { getAppConfig } from './config.js';
+
+const config = getAppConfig();
+
+const logger = createLogger({
+	logFile: !config.isProduction,
```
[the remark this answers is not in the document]
joneubank: We've had issues with pods failing silently as a result of log file writes so stopping them in any deployed environment is important.

mistryrn (approved):
🚀

Issue #154, opened 2023-10-13
Create Wizard Progress Bar + Title


b-f-chan:
Create reusable UI components for the wizard progress bar + title

**Expected Outcome(s)**

- [x] `Back to Dashboard` link at top is available on the screen
- [x] Title at top of page exists
- [x] White container with the wizard body exists
- [x] Progress bar with div exists
- [x] Different visual states are implemented (step incomplete, complete, etc)

alyssadsouza:
@b-f-chan @patrickdsantos here's the QA link: https://consent-ui.dev.ohcrn.cancercollaboratory.org/en/consent/step-2 and just a few notes:

- You can navigate between the different steps with the URL (like `/step-1` or `/step-3`) since there's no links to next pages right now
- The progress of the user is set to step 1: complete, steps 2-5: incomplete
- The "In Progress" state is for shown for whatever the current step of the page is, so if you're on https://consent-ui.dev.ohcrn.cancercollaboratory.org/en/consent/step-1 it will show step 1 as in progress

Let me know if you have any questions about that!

patrickdsantos:
Amazing work @alyssadsouza , the progress bar looks fantastic on all device sizes! Only one piece of feedback and not sure if this is something that's exclusive to staging but the line between step 1 and 2 is green even when step 1 is set to in progress. 
<img width="525" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVYrQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2461e5c3b06991a05f0877e8917158ae215f8bd8/Screenshot%202023-12-21%20at%2010.31.25%20AM.png" alt="Screenshot 2023-12-21 at 10.31.25 AM.png" />

patrickdsantos:
Disregard the feedback! It is indeed a staging thing, no changes needed. Thanks again @alyssadsouza !!

b-f-chan:
Done, closing

PR #274, opened 2023-11-21
💄📱 #154: Add Consent Wizard layout

alyssadsouza:
## Description of Changes

Adds the Consent Wizard layout, which includes a reusable `PaddedContainer` component and `Back to Dashboard` link header component (that will be reused on multiple pages) in addition to a reusable `ProgressBarHeader` component.

### Consent UI

**Consent Wizard Layout**
- Replaced the `ConsentForm` component with `ConsentWizard` to act as a container for the entire Consent Wizard
- Added the `PaddedContainer` component to be reused on most pages with the Card layout which has the 15px, 20px, 50px padding for mobile, tablet, and desktop screens respectively
- Added the `BackToDashboard` header component which has the back to dashboard link on top
- Added the reusable `ProgressHeader` component which is used in Consent Wizard but can display progress for any number/list of steps

**Other changes**
- Updated the `DashboardComponent` to use the same `PaddedContainer` component for the layout surrounding the consents card
- Abstracted the common Link styling (underlined, teal colour, hover state) on `LocalizedLink` components away to its own style module, and added the `defaultStyle` prop to allow disabling the default styles
- Removed the `padding` in the `Card` component and added it in the specific components styles (`Dashboard`, `Modal`, `LandingPageCard`) since it differs across the dashboard, landing page, etc and it was getting hard to override the padding defined in `Card`'s styles

>**Note about padding vs. margins**: The `PaddedContainer` uses padding (as the name suggests) as opposed to margins. This is because when shrinking to smaller screen widths, the padding is maintained whereas margins allow for the inner content to take up the entire screen width, which basically means dissolving the left and right margins like this:
>Screenshot 2023-11-30 at 11 52 42 AM

**Preview**

Screenshot 2023-11-30 at 11 56 05 AM

**Responsive screens**

https://github.com/OHCRN/platform/assets/70766739/99a10908-1431-4097-900d-5c7648808fb5

## PR Readiness Checklist

- [x] PR feedback has been addressed
- [x] PR branch is up-to-date with target branch and there are no merge conflicts
- [x] Connected ticket to PR
- [x] Labels added to PR


apps/consent-ui/src/components/views/ConsentWizard/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+			inProgress: step === currentStep,
+		};
+	});
+
+	return (
+		<PaddedContainer>
+			<BackToDashboard currentLang={currentLang} />
+			<div className={styles.wizard}>
+				<div className={styles.header}>
+					<h3>{translate('consentWizard', 'heading')}</h3>
```
mistryrn, line 72: Based on the structure of this page, I think we want this to be our `h1`

apps/consent-ui/src/components/common/ProgressHeader/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+					<>
+						<div className={styles.step} key={step.name}>
+							{step.inProgress ? (
+								<InProgress className={styles['in-progress']} />
+							) : step.isComplete ? (
+								<Success className={styles.success} />
+							) : (
+								<Incomplete className={styles.incomplete} />
+							)}
+							<h3 className={styles['desktop-step-name']}>{step.name}</h3>
```
mistryrn, line 58: For accessibility reasons, I'm not sure section heading tags are the best choice inside the progress tracker. My reasoning being that this progress tracker doesn't contain any content, just displays the state/progress of the consent. There aren't any sections of content between each step in the tracker, if that makes sense?

A common use case for heading tags with screen readers is to scan the HTML and pull up a list of the headings, and then the user can navigate directly to a specific heading. With all of these as headings, it might be a bit confusing if someone navigated to `2. Consent to Release Data` and then the next element read out to them is the progress of the next step.

mistryrn (commented):
Awesome stuff. Tested locally and it's looking great! Thanks for calling out where we'll need to make a Consent API call to get the actual progress once implemented. Should be pretty straight forward to hook this up down the line :)

Just had a few comments about the heading tags, other than that it looks good to me! Let's wait until Sam's Forms PR #235 is merged before merging this though

apps/consent-ui/src/components/common/ProgressHeader/index.tsx
alyssadsouza: yeah, makes sense!

apps/consent-ui/src/components/views/Dashboard/Dashboard.module.scss
```diff
@@ -1,40 +1,35 @@
-.dashboard {
```
alyssadsouza, line 1: removing the dashboard class because the `PaddedContainer` contains the necessary styles

apps/consent-ui/src/components/views/Home/LandingPageCard/LandingPageCard.module.scss
```diff
@@ -1,16 +1,12 @@
 .landing-page-card {
-	@media (min-width: theme('screens.md')) {
```
alyssadsouza, line 2: removing the link styling and using the abstracted link styles for those links

mistryrn (approved):
Thanks for getting those changes in! Tested locally and it still works great 🚀

samrichca (approved):

Issue #275, opened 2023-11-21
[Backend] Implement Logging, Error Handling, Validation Best Practices for Clinician Invite Flow


b-f-chan:
Implement logging, error handling, validation best practices across all backend components for the E2E clinician invite flow.  These components include:

- [x] DAS
- [x] Data Mapper
- [x] Consent API

# Expected Outcome(s):

The following mechanisms are reviewable in each component:

- [x] Logging
- [x] Error handling
- [x] Validation

Please reference @joneubank's work establishing these approaches

b-f-chan:
Done, closing

Issue #168, opened 2023-10-24
Build Consent API Endpoint --> Clinician Invite Page: `POST /invites`


b-f-chan:
Build consent API endpoint: `POST /invites`

**NOTE:** This should send the form data and recaptcha token

**Applicable Page:** Clinician Invite
**Functionality:** Clinician enters Participant details and sends invitation

**Expected Outcome(s):**

- [x] Swagger doc for the endpoint is functional and testable
- [x] Unit testing implemented for this endpoint

---

Update 10/30/2023 from Sam:

- Endpoint mapped out here https://github.com/OHCRN/platform/pull/191

b-f-chan:
Need to do mapping as pre-requisite, moving to backlog

b-f-chan:
Merged, closing

PR #279, opened 2023-11-22
✨ POC: Build `POST /invites` route in Consent API (#168)

alyssadsouza:
## Description of Changes

Implemented the ~~"happy path"~~ `POST /invites` route in Consent API which integrates with Data Mapper and Consent DAS + PI DAS to create a clinician invite. Also added a POC for unit testing on the `routers`.

### Consent API
Added the `POST /invites` route with incomplete error handling (E2E error handling to be considered in #275).
- Renamed the `participantResponses.ts` service file to `create.ts` and added the `createInvite` service which makes a POST request to Data Mapper's `POST /invites` route
- Created base `axiosClient` instance to reuse for API calls in Consent API
- Added a test suite for the endpoint in `tests/routers/invites.ts` as a POC, using `supertest` to make requests to our endpoints through the `App` module, and [mocking](https://vitest.dev/guide/mocking.html#modules) the `createInvite` function which makes an `API` call to `data-mapper`
- Added package [`url-join`](https://www.npmjs.com/package/url-join) to format API request path names
- Added package [`supertest`](https://github.com/ladjs/supertest) to make requests to the router by passing in the `App` module, allowing us write unit tests for the endpoints

### Types
- Renamed `httpErrors` dir to `httpResponses`
- Added `Results.ts` file to include types that we use for returning success or failure responses in services, based on example provided by Jon here https://gist.github.com/joneubank/ac5a5b30dafb84915d633131c54aa028
- Added type `ConflictErrorResponse` that extends `ErrorResponse` and returns a `CONFLICT_ERROR` for the `error` field
- Added an `ErrorName` enum to specify accepted `error` strings
- Updated the `RequestValidationErrorResponse` type  to return Zod `issues` rather than `flatten()` the error details, so we get details for nested objects

### Special Instructions
Before running these changes, you will need to install `url-join` and `supertest`:
```
pnpm i
```

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-api/tests/services/create.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
+		const invite = await createInvite(mockInviteRequest);
+
+		expect(invite.success).toBe(true);
+		if (invite.success) {
+			const response = invite.data;
+			expect(response).toContain(mockInviteRequest);
+			expect(response.id).toBeDefined();
+			expect(response.inviteAccepted).toBe(false);
+			expect(response.inviteSentDate).toBeDefined();
```
alyssadsouza, line 53: This test case is pretty trivial even if it's just meant to be the groundwork for unit testing... although what it does do that the Zod schemas can't do is verify that the response body matches all the fields that were present in the request body exactly (even though that is a very obvious expected behaviour)

apps/consent-api/src/service/axiosClient.ts
alyssadsouza, line 1: The `service` dir will need to be renamed to `services`, I'm guessing in a new PR

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -19,9 +19,9 @@
 
 import axios from 'axios';
 import { RequestHandler } from 'express';
+import { ErrorResponse } from 'types/httpErrors';
```
joneubank, line 22: This is evidence I broke the linting rules. I will retire in shame.

apps/consent-api/tests/services/create.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+describe('POST /invites', () => {
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
+		mockCreateInvite();
+		const appConfig = getAppConfig();
+		const response = await request(App(appConfig)).post('/invites').send({
+			recaptchaToken: '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe', // test recaptchaToken key
```
alyssadsouza, line 69: the tests are failing because this token isn't valid when tests are run in the root dir, though it passes when running tests in `consent-api`

apps/consent-api/tests/routers/invites.ts
```diff
@@ -0,0 +1,94 @@
...
+describe('POST /invites', () => {
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
+		mockCreateInvite();
+		const appConfig = getAppConfig();
+		const response = await request(App(appConfig)).post('/invites').send({
+			recaptchaToken: '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe', // test recaptchaToken key
```
alyssadsouza, line 69: the tests are failing because this token isn't valid when tests are run in the root dir, though it passes when running tests in consent-api

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+describe('POST /invites', () => {
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
+		mockCreateInvite();
+		const appConfig = getAppConfig();
+		const response = await request(App(appConfig)).post('/invites').send({
+			recaptchaToken: '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe', // test recaptchaToken key
```
alyssadsouza, line 69: the tests are failing because this token isn't valid when tests are run in the root dir, though it passes when running tests in consent-api

alyssadsouza: using [`stubEnv`](https://vitest.dev/api/vi#vi-stubenv) to set `NODE_ENV=development` for now to pass tests (setting the `RECAPTCHA_SECRET_KEY` isn't working for some reason), will remove later when the env vars setup for Jenkins is figured out

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,91 @@
...
+	guardianPhoneNumber,
+	guardianEmailAddress,
+	guardianRelationship,
+	clinicianFirstName,
+	clinicianLastName,
+	clinicianInstitutionalEmailAddress,
+	clinicianTitleOrRole,
+	consentGroup,
+	consentToBeContacted,
+}: ClinicianInviteRequest): Promise<
```
joneubank, line 69: Why are we deconstructing all this instead of just passing the entire object?

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,91 @@
...
+}): Promise<any> => {
+	const updateObj = {
+		consentQuestionId,
+		participantId,
+		response,
+	};
+
+	// TODO: POST to data-mapper, then return the result
+	return updateObj;
+};
```
joneubank, line 44: Should not be returning `any`, if you are defining a stub with a TODO let's at least specify the inputs/outputs accurately.

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,91 @@
...
+		guardianEmailAddress,
+		guardianRelationship,
+		clinicianFirstName,
+		clinicianLastName,
+		clinicianInstitutionalEmailAddress,
+		clinicianTitleOrRole,
+		consentGroup,
+		consentToBeContacted,
+	});
+	return ClinicianInviteResponse.safeParse(data);
```
joneubank, line 90: We also need to handle failed requests. Should have a try catch around axios requests to handle unexpected thrown errors, and also have a check for non 200 response codes (axios responses have a `.ok`).

It is good that we are checking for invalid response structures and I really like that we are returning a type that includes a success/failure boolean, but I don't think we need to return the validation info (we just log it so that a developer can fix the problem). When another system is returning invalid responses our service should return an error 500 with a generic "Unexpected server error" message.

There is a really good pattern for functions like this, where on success you return an object of type `{success:true, data: ClinicianInviteResponse}` and on failure you can return `{success:false, message: string}`. We could create a generic type for this response so we don't need to return the Zod parse error, here's a gist I have of a `Result<T>` type that accomplishes this: https://gist.github.com/joneubank/ac5a5b30dafb84915d633131c54aa028

apps/consent-api/src/routers/invites.ts
```diff
@@ -71,9 +70,27 @@ router.post(
...
+				// ClinicianInviteResponse failed validation from Data Mapper even though the POST request was successful
+				return res
+					.status(500)
+					.send(
+						ResponseValidationErrorResponse(
+							invite.error,
+							'ClinicianInviteResponse from Data Mapper was malformed',
+						),
+					);
+			}
```
joneubank, line 87: Repeating my previous comment... We don't need to tell the client about failures from our backend systems. We should just return a generic server error message and log the problem for us to resolve.

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+	};
+	return { inviteRequest, inviteResponse };
+});
+
+vi.mock('../../src/services/create.js', () => {
+	// mock the createInvite service so we don't need to make an API call to data-mapper
+	return { createInvite: () => ClinicianInviteResponse.safeParse(mocks.inviteResponse) };
+});
+// TODO: can remove when env vars setup in jenkins is figured out, find out why setting RECAPTCHA_SECRET_KEY wont work
+vi.stubEnv('NODE_ENV', 'development');
```
joneubank, line 63: Perhaps we can have a common file to setup all tests, I imagine there are other tests that will need this.

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+describe('POST /invites', () => {
+	afterAll(() => {
+		vi.restoreAllMocks();
```
joneubank, line 67: 👍

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+		vi.unstubAllEnvs();
```
joneubank, line 68: Is this required after each test? Or just unstub at the end? I don't know what vitest recommends.

docs/shared-types.md
```diff
@@ -46,6 +46,8 @@ Both of these convert the schema to a `ZodEffects` type which "is a wrapper clas
 
 Zod's [`parse()`](https://zod.dev/?id=parse) method simultaneously validates the data and can also coerce any data types or transform fields, as outlined in the Zod schema.
 
+>**Note**: Zod's `parse` does **not** throw an error if it encounters unrecognized fields, it simply parses those out of the output. To disallow unknown fields, see [`strict()`](https://zod.dev/?id=strict).
```
joneubank, line 49: Thanks for including a reference link with this note :)

packages/types/src/httpErrors/RequestValidationErrorResponse.ts
```diff
@@ -17,14 +17,14 @@
...
+import { ZodError, ZodIssue } from 'zod';
 
 import { ErrorResponse } from './ErrorResponse.js';
 
 export const REQUEST_VALIDATION_ERROR = 'RequestValidationError';
 
-export type RequestValidationError<T> = ErrorResponse & {
-	details: typeToFlattenedError<T>;
+export type RequestValidationError = ErrorResponse & {
+	details: ZodIssue[];
```
joneubank, line 27: 🙏

joneubank (changes requested):
Requesting changes to the error handling written around the axios request to datamapper.

apps/consent-api/src/services/create.ts
alyssadsouza: Ah good point, no reason I'll change that

apps/consent-api/src/services/create.ts
alyssadsouza: This was just code I moved over from `services/participantResponses.ts` to `services/create.ts` - not sure if I should delete it since it's an old implementation and needs to be updated anyway (which it will when the consent question responses endpoint is built)

apps/consent-api/src/services/create.ts
alyssadsouza: Thanks for the gist! I don't see an `ok` field, there is a `statusText` field https://axios-http.com/docs/res_schema which returns the HTTP message for the status code (e.g. `200: "OK"`, `201: "Created"`) so in this case we'd be looking for `statusText === "Created"`. Maybe it's better to just validate the status code is a 201 as we explicitly set in the `data-mapper` endpoint?

The other thing I'm wondering about the `Result` type is if it includes all the information we want. The bad request example I run into a lot is where the participant email I'm creating an invite with is not unique (which is a constraint on it in the `PI` database), so there already exists an invite with that same email address, and that cannot be validated (at least not yet) until we try to create an invite in PI DAS and get a Prisma error. I think that would classify as a 400 error, so I'm wondering what's the best way to return that information from `createInvite` to the router so that it gives the correct error.

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,85 @@
...
+		const invite = ClinicianInviteResponse.safeParse(data);
+		if (!invite.success) {
+			logger.error(`[Data-Mapper] 😱 'POST /invites' had an invalid response`, invite.error.issues);
+			return failure('');
+		}
+		return success(invite.data);
+	} catch (error) {
+		if (error instanceof AxiosError) {
+			// TODO: E2E error handling → this could be any error from requests to data-mapper, pi-das or consent-das
+			return failure(error.message);
```
alyssadsouza, line 81: Not sure how this should be treating errors that might be more than server errors, for example: "an invite with that participant email already exists", because I think ideally we would want to return a 400 response in the router.

apps/consent-api/tests/routers/invites.spec.ts
alyssadsouza: From the docs https://vitest.dev/api/vi.html#vi-unstuballenvs:

> Restores all import.meta.env and process.env values that were changed with vi.stubEnv. When it's called for the first time, Vitest remembers the original value and will store it, until unstubAllEnvs is called again.

I _think_ that means each use will just restore the previous value? In which case we would only want to call it at the end.

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,85 @@
...
+		const { data, status } = await axiosClient.post(
+			urlJoin(dataMapperUrl, 'invites'),
+			inviteRequest,
+		);
+
+		if (status !== 201) {
+			// received a non 201 response but didn't get caught as an error
+			logger.error(`[Data-Mapper] 'POST /invites' received ${status} status`);
+			return failure('');
+		}
```
joneubank, line 70: I don't think 201 is necessarily an error condition - so long as we have received the data we need in the body we can validate it. We can accept this response and pass teh data to our validator below... if it is an empty body or fails validation then we handle it there.

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,85 @@
...
+
+		const invite = ClinicianInviteResponse.safeParse(data);
+		if (!invite.success) {
+			logger.error(`[Data-Mapper] 😱 'POST /invites' had an invalid response`, invite.error.issues);
```
joneubank, line 74: emoji is cute and made me laugh... I'm going to ask for us to keep log messages simple and dry though, sorry :( Can we remove the emoji?

I would like to recommend another approach than this `[Data-Mapper]` tag here. As is, it could be interpreted as the error being thrown within the data-mapper application. What we would like to express is that this log message comes from the dataMapperClient. I added a mechanism to do this to the logger:

```ts
import serviceLogger from '../logger.js';
const logger = serviceLogger.forModule('DataMapperClient');
// ...
logger.error('POST /invites', 'Received invalid data in response.', invite.error.issues);
// "[Consent-API.DataMapperClient] POST /invites - Received invalid data in response. - { ... issues ... }
```

apps/consent-api/src/services/create.ts
joneubank: Yes, interesting use case. Since we have multiple response statuses now, not just Success and Failure, we can create a more descriptive type to respond with that has multiple statuses:

```ts
type CreateInviteResult =
	| { status: 'SUCCESS'; data: CreateInviteResponse }
	| { status: 'SYSTEM_ERROR'; message: string }
	| { status: 'INVITE_EXISTS'; message: string };

export const createInvite = async (
	inviteRequest: ClinicianInviteRequest,
): Promise<CreateInviteResult> =>
```

So now when we get the result back we can handle these cases in a switch:

```ts
const inviteResult = await createInvite(req.body.data);
switch(inviteResult.status) {
  case 'SUCCESS': {
    // validate inviteResult.data
    return res....;
  }
  case 'SYSTEM_ERROR': {
    return res.status(500).json(ErrorResponse(inviteResult.message);
  }
  case 'INVITE_EXISTS': {
    return res.status(400).json(ErrorResponse(inviteResult.message);
  }
```

The advantage of structuring our code this way is that our `createInvite` code now details all the different cases that our service may need to deal with. All our code downstream doesn't know about the HTTP communication and how that API works, it only knows if the invite was created, and if not created it has a list of reasons why.

Its worth noting: I don't know if there is an error sending a duplicate invitation... double check with Rakesh and Brandon how the UX is supposed to work in that case.

apps/consent-api/tests/routers/invites.spec.ts
joneubank: sounds useful for afterAll. Thanks for the reference!

joneubank (commented):
Responded to your question re: existing invite. Good question.

Also something about logger formatting.

apps/consent-api/src/services/create.ts
alyssadsouza: Still waiting on word about that from Brandon, but I updated the `Result` type to use `status` to indicate success or failure as opposed to the `success` boolean, since it allows us to structure our error handling like what you suggested. If it's overcomplicating things right now then I'm open to removing that type from the shared `types` package altogether and just going with the local `CreateInviteResult` type you defined above, but I think it's useful to have the global `Result` type with the `success()`, `failure()` and `alternate()` methods you defined in the gist so we can use them in other routers and have consistency in the services response types.

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,81 @@
...
+
+		if (!invite.success) {
+			logger.error('POST /invites', 'Received invalid data in response.', invite.error.issues);
+			return failure('SYSTEM_ERROR', invite.error.message);
+		}
+
+		return success(invite.data);
+	} catch (error) {
+		if (error instanceof AxiosError) {
+			// TODO: E2E error handling → this could be any error from requests to data-mapper, pi-das or consent-das
```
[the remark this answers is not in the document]
alyssadsouza: Yeah that's right, `error.response.data` returns whatever our error response was.

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -71,6 +73,6 @@ export const recaptchaMiddleware: RequestHandler = async (req, res, next) => {
 		next();
 	} else {
 		// TODO: Need a formatted error for reCAPTCHA validation failures
```
anncatton, line 75: does this `TODO` apply now that the error is in place? or is it referring to something else 🤔

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -71,6 +73,6 @@ export const recaptchaMiddleware: RequestHandler = async (req, res, next) => {
 		next();
```
anncatton, line 73: can add `return`s here, like:
``` if (recaptchaVerified) {
		return next();
}
return res.status(400).json(ErrorResponse(RECAPTCHA_ERROR, 'Valid reCAPTCHA token is required.'));
```
can get rid of the `else` too

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -19,9 +19,11 @@
...
 import axios from 'axios';
 import { RequestHandler } from 'express';
+import { ErrorName, ErrorResponse } from 'types/httpErrors';
+
+const { RECAPTCHA_ERROR } = ErrorName;
 
 import { getAppConfig } from '../config.js';
-import { ErrorResponse } from 'types/httpErrors';
 
 const verifyRecaptcha = async (recaptchaToken?: string | null) => {
```
anncatton, line 28: while in this file can you update the `console.log()`s to `logger` statements?

packages/types/src/httpErrors/ErrorResponse.ts
```diff
@@ -17,9 +17,19 @@
...
+export enum ErrorName {
+	SERVER_ERROR = 'ServerError',
+	CONFLICT_ERROR = 'ConflictError',
+	REQUEST_VALIDATION_ERROR = 'RequestValidationError',
+	RECAPTCHA_ERROR = 'RecaptchaError',
+}
+
 export type ErrorResponse = {
-	error: string;
+	error: ErrorName | 'NOT_IMPLEMENTED'; // TODO: remove once all routes are implemented
```
anncatton, line 28: 👍

packages/types/src/httpErrors/ConflictErrorResponse.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Creates a ConflictErrorResponse containing the target fields causing the conflict and a message detailing it.
+ * @param target list of target fields that are causing the conflict e.g. `['participantEmailAddress']`
+ * @param customMessage
+ * @returns
+ */
+export const ConflictErrorResponse = (target: string[], customMessage?: string): ConflictError => ({
+	error: CONFLICT_ERROR,
+	message: customMessage ?? 'There was a conflict with existing data.',
+	target,
```
anncatton, line 37: curious about `target` as a key name...i think makes more sense to call it `fields` or `onFields`...what do you think?

packages/types/src/httpErrors/ConflictErrorResponse.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ErrorName, ErrorResponse } from './ErrorResponse.js';
+
+const { CONFLICT_ERROR } = ErrorName;
+
+export type ConflictError = ErrorResponse & {
+	target: string[];
+};
+
+/**
+ * Creates a ConflictErrorResponse containing the target fields causing the conflict and a message detailing it.
```
anncatton, line 29: thanks for the documentation!

apps/consent-api/tests/routers/invites.spec.ts
anncatton: For the env config, should we just stub the entire app config instead? so no surprises

apps/consent-api/src/services/create.ts
anncatton: Can we leave it as is for now? Since it's not fully implemented but this request is referenced elsewhere iirc, but isn't part of the work for this PR.

packages/types/src/httpErrors/Result.ts
```diff
@@ -0,0 +1,65 @@
+// Success and Failure types
+export type Success<T> = { status: 'SUCCESS'; data: T };
+export type Failure<T = void, FailureStatus = string> = {
+	status: FailureStatus;
+	message: string;
+	target?: string[]; // list of target fields pertaining to the error, for example `emailAddress` for a conflict error
```
anncatton, line 6: instead of adding the `target` key to the `Failure` type, i wonder if we can extend this type when this is applicable, so that `Failure` remains more basic in structure. 🤔

packages/types/src/httpErrors/Result.ts
```diff
@@ -0,0 +1,65 @@
...
+/* ******************* *
+   Convenience Methods 
+ * ******************* */
+
+/**
+ * Create a successful response for a Result or Either type, with data of the success type
+ * @param {T} data
+ * @returns {Success<T>} `{status: 'SUCCESS', data}`
+ */
+export const success = <T>(data: T): Success<T> => ({ status: 'SUCCESS', data });
```
mistryrn, line 32: Bit of a nitpick, but if we're going to include `success`, should we rename this from `httpErrors` to `httpResponses`?

packages/types/src/httpErrors/ErrorResponse.ts
mistryrn, line 28: 👍

mistryrn (commented):
Awesome work! Just a few small requests 🙂

packages/types/src/httpErrors/ConflictErrorResponse.ts
alyssadsouza: Yeah, I got `target` from the name that `PrismaClientKnownRequestError` gives to the field that caused the error https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientknownrequesterror, but since this is specific to a conflict error something like `onFields` is probably more clear.

packages/types/src/httpErrors/Result.ts
alyssadsouza: I was thinking that too, I'll do that

packages/types/src/httpErrors/Result.ts
alyssadsouza: good idea! added a `Conflict` type to `Results.ts` with the assumption we will use it more than once

packages/types/src/httpErrors/ConflictErrorResponse.ts
alyssadsouza: renamed to `onFields`

mistryrn (commented):
New changes look great, thanks for getting those updates in so quickly! I think Ann had one or two outstanding bits of feedback that she's thinking through, so I'll wait for that before approving, but other than that I think this is good to go 😄

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,84 @@
...
+});
+
+/**
+ * Create a response indicating a conflict with a status naming the reason and message describing the conflict.
+ * @param status status name
+ * @param onFields list of field names causing the conflict
+ * @param message message detailing the conflict
+ * @returns {Conflict} `{status: string, message: string, data: undefined, onFields: string[]}`
+ */
+export const conflict = <ConflictStatus>(
```
anncatton, line 60: from zoom call: we can remove this extended type and use `failure` with `INVITE_EXISTS` with specifics on the conflict coming from the message

packages/types/src/httpResponses/ConflictErrorResponse.ts
```diff
@@ -0,0 +1,41 @@
...
+	onFields: string[];
+};
+
+/**
+ * Creates a ConflictErrorResponse containing the target fields causing the conflict and a message detailing it.
+ * @param onFields list of onFields fields that are causing the conflict e.g. `['participantEmailAddress']`
+ * @param customMessage
+ * @returns
+ */
+export const ConflictErrorResponse = (
```
anncatton, line 34: i'm still happy to keep this extended type, it's a nice convenience and is extra readable. but we can remove the `onFields` property and rely on the information in the message.

apps/consent-api/src/routers/invites.ts
```diff
@@ -71,9 +73,23 @@ router.post(
...
+					return res.status(201).send(invite.data);
+				}
+				case 'SYSTEM_ERROR': {
+					return res.status(500).json(ErrorResponse(SERVER_ERROR, invite.message));
+				}
+				case 'INVITE_EXISTS': {
+					return res.status(409).json(ConflictErrorResponse(invite.onFields, invite.message));
+				}
+			}
+		} catch (error) {
```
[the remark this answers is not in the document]
alyssadsouza: Actually no, because if none of the cases are caught the code will just resume after the `switch` statement, i.e. end there. It wouldn't be caught in the `catch` block because no error would be thrown, the `try` block just completes and the function returns nothing. We don't necessarily *need* an explicit default case if we add a final return statement at the end of the `switch` statement, which will just run if none of the other cases are caught, so basically acting as a default. I'll just add that it in and have it return a 500 response with an "unexpected error" message.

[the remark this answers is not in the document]
alyssadsouza: Update: TS catches any code under that `switch` as unreachable code because it can tell that all the status cases were caught 😛
Screenshot 2023-11-29 at 11 28 44 AM
And nice to see it works if I remove a case:
Screenshot 2023-11-29 at 11 29 01 AM
So... I'll just leave it as is.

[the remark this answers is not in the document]
anncatton: Nice! Thanks for verifying this 😀

docs/error-handling.md
```diff
@@ -0,0 +1,137 @@
...
+| -------------------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `SERVER_ERROR`             | 500    | An issue occurred on our server side. When another system is returning invalid responses, our service should return a 500 status with a generic "Unexpected server error" message. |
+| `CONFLICT_ERROR`           | 409    | A request is prevented from being made because it conflicts with existing data.                                                                                                    |
+| `REQUEST_VALIDATION_ERROR` | 400    | A request was made with an invalid request body.                                                                                                                                   |
+| `RECAPTCHA_ERROR`          | 400    | A ReCAPTCHA token could not be verified.                                                                                                                                           |
```
anncatton, line 73: thanks for this! 😎

docs/error-handling.md
mistryrn, line 1: This is fantastic 👏

mistryrn (approved):
Great work on this 👏 Looks like all of mine and @anncatton 's feedback has been addressed so I'm approving, but I think we still need approval from @joneubank before you can merge since he requested changes previously

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const createResponse = async ({
+	consentQuestionId,
+	participantId,
+	response,
+}: {
+	consentQuestionId: string;
+	participantId: string;
+	response: boolean;
+}): Promise<any> => {
```
joneubank, line 40: No `any` responses! If you know the type you want to return, then use that so TS can validate you are returning the correct type. Otherwise, don't declare return type and let TS determine what type the function returns.

This function looks to be a placeholder for future dev, but let's not put the wrong types in it (in fear that they get used in the future).

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,60 @@
+// Success and Failure types
```
joneubank, line 1: File needs copyright header.

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,60 @@
...
+	status: FailureStatus;
+	message: string;
+	data: T;
+};
+
+/**
+ * Represents a response that on success will include data of type A,
+ * and on failure will return data of type B
+ */
+export type Either<A, B> = Success<A> | Failure<B>;
```
joneubank, line 13: With how you have refactored the Result to have a FailureStatus generic, we no longer need `Either`. Can remove this.

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const alternate = <T, FailureStatus>(
+	status: FailureStatus,
+	data: T,
+	message: string,
+): Failure<T, FailureStatus> => ({
+	status,
+	data,
+	message,
+});
```
joneubank, line 60: Can remove this now that `failure` takes a generic

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,60 @@
+// Success and Failure types
+export type Success<T> = { status: 'SUCCESS'; data: T };
+export type Failure<T = void, FailureStatus = string> = {
+	status: FailureStatus;
+	message: string;
+	data: T;
+};
```
joneubank, line 7: I like how this has been extended to include multiple failure statuses. I would clean it up a bit by reworking it as:
```ts
// Success and Failure types
export type Success<T> = { status: 'SUCCESS'; data: T };
export type Failure<FailureStatus extends string, T = void> = {
	status: FailureStatus;
	message: string;
	data: T;
};

/**
 * Represents a response that on success will include data of type T,
 * otherwise a message will be returned in place of the data explaining the failure.
 * The failure object has data type of void.
 */
export type Result<T, FailureStatus extends string, FailureData = void> =
	| Success<T>
	| Failure<FailureStatus>;
```

This way we don't need the `Either` type, as we can include that in the HttpFailure. That also means we can remove the `alternate` method. These were awkwardly named anyways.

Using the `extends string` with the FailureStatus generic means that the implementer must state which failure cases could be returned. It could look like:

```ts
type DoActionResult = Result<{data: string}, 'INVITE_EXISTS' | 'SYSTEM_ERROR'>;
```

If you also want data to be returned with the error:

```ts
type DoActionResult = Result<{data: string}, 'INVITE_EXISTS' | 'SYSTEM_ERROR', {failureData: number}>;
```

packages/types/src/httpResponses/Result.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns {Failure} `{status: string, message: string, data: undefined}`
+ */
+export const failure = <FailureStatus>(
+	status: FailureStatus,
+	message: string,
+): Failure<void, FailureStatus> => ({
+	status,
+	message,
+	data: undefined,
+});
```
joneubank, line 45: Will need a generic for the return data type.

apps/consent-api/tests/config.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { vi } from 'vitest';
+
+export const mockEnv = () => {
+	vi.stubEnv('NODE_ENV', 'development');
+	vi.stubEnv('PORT', '8080');
+	vi.stubEnv('RECAPTCHA_SECRET_KEY', '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe'); // SECRET TEST KEY
+	vi.stubEnv('DATA_MAPPER_URL', 'http://localhost:8081');
+};
```
joneubank, line 27: This is a really good idea, but I might suggest doing this a bit differently... We realyl want to make sure our `getAppConfig` function from the config module is returning a fully configured `AppConfig` object, and we may want to allow individual test suites to modify the env returned.  Here is an alternative approach I would recommend:

```ts
import * as config from '../src/config.js';
import { AppConfig } from '../src/config.js';

export const mockEnv = (values?: Partial<AppConfig>) => {
	// Need to set the NODE_ENV since there are currently places in teh code checking the process.env.NODE_ENV directly instead of relying on our config.
	vi.stubEnv('NODE_ENV', values?.isProduction ? 'production' : 'development');

	const defaultConfig: AppConfig = {
		dataMapperUrl: 'http://localhost:8081',
		express: {
			port: '8080',
		},
		isProduction: false,
		recaptcha: {
			secretKey: '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe',
		},
	};

	vi.spyOn(config, 'getAppConfig').mockReturnValue({ ...defaultConfig, ...values });
};
```

This will ensure that we always have the full `AppConfig` in the mock, since TS will yell at us if we are missing any default properties. We can also now provide replacement values for any test suite.

One added benefit, it will catch code that is reading directly from the `process.env` in stead of relying on our config module (since we aren't adding mock env values, other than `NODE_ENV` which is a standard).

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+	beforeAll(() => mockEnv());
+	afterAll(() => {
+		vi.restoreAllMocks();
+		vi.unstubAllEnvs();
+	});
+
+	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
+		const appConfig = getAppConfig();
+		const response = await request(App(appConfig)).post('/invites').send({
+			data: mocks.inviteRequest,
```
joneubank, line 78: 👍

apps/consent-api/src/services/create.ts
```diff
@@ -0,0 +1,88 @@
...
+		return success(invite.data);
+	} catch (error) {
+		if (error instanceof AxiosError && error.response) {
+			// TODO: E2E error handling → this could be any error from requests to data-mapper, pi-das or consent-das
+			const { data, status } = error.response;
+			logger.error('POST /invites', 'AxiosError handling create invite request', data);
+
+			if (status === 409) {
+				return failure('INVITE_EXISTS', data.error);
+			}
```
joneubank, line 81: 👍

apps/consent-api/src/routers/invites.ts
```diff
@@ -60,20 +64,36 @@ const ClinicianInviteSchema = z.object({ data: ClinicianInviteRequest });
...
+				case 'SUCCESS': {
+					return res.status(201).send(invite.data);
+				}
+				case 'SYSTEM_ERROR': {
+					return res.status(500).json(ErrorResponse(SERVER_ERROR, invite.message));
+				}
+				case 'INVITE_EXISTS': {
+					return res.status(409).json(ConflictErrorResponse(invite.message));
+				}
+			}
```
joneubank, line 92: ❤️

apps/consent-api/src/routers/invites.ts
```diff
@@ -60,20 +64,36 @@ const ClinicianInviteSchema = z.object({ data: ClinicianInviteRequest });
...
 	recaptchaMiddleware,
 	withRequestBodyValidation(ClinicianInviteSchema, async (req, res) => {
-		return res
-			.status(500)
-			.send(ErrorResponse('NOT_IMPLEMENTED', 'Route has not been implemented.'));
+		try {
+			const invite = await createInvite(req.body.data);
+			switch (invite.status) {
+				case 'SUCCESS': {
+					return res.status(201).send(invite.data);
```
joneubank, line 84: Could this also use `.json`?

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -35,10 +40,10 @@ const verifyRecaptcha = async (recaptchaToken?: string | null) => {
 		const recaptchaVerification = await axios.post(
 			`https://www.google.com/recaptcha/api/siteverify?secret=${config.recaptcha.secretKey}&response=${recaptchaToken}`,
 		);
-		console.log('success?');
+		logger.debug('Successful ReCAPTCHA verification');
```
joneubank, line 43: This is going to get real noisy without giving us much useful info to work with, but we can keep it until we get annoyed.

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -19,9 +19,14 @@
 
 import axios from 'axios';
 import { RequestHandler } from 'express';
+import { ErrorName, ErrorResponse } from 'types/httpResponses';
 
 import { getAppConfig } from '../config.js';
-import { ErrorResponse } from 'types/httpErrors';
+import serviceLogger from '../logger.js';
+
+const logger = serviceLogger.forModule('ReCAPTCHAVerification');
```
joneubank, line 27: 🙏

joneubank (changes requested):
There is so much good stuff in here, thanks a million. If it seems like I'm being picky or refactoring too much it is because the PR is good and that makes it engaging, also since this is going to be a template for future work I do want to push our standards a bit.

Couple suggestions for rework here, honestly not much left to cleanup.

I love all the documentation added. Thats huge.

packages/types/src/httpResponses/Result.ts
alyssadsouza: Ah that's right! Thanks for catching that

packages/types/src/httpResponses/Result.ts
alyssadsouza: It looks like the `alternate` type is needed because TS has a fit if you try to make generic typed function params optional so I'm leaving it in because I think it's more convenient than having to explicitly pass `undefined` to `failure` when I don't want to give it data

joneubank (approved):

PR #293, opened 2023-12-04
🦺 PI DAS error handling for Clinician Invite flow (#275)

alyssadsouza:
## Description of Changes

Implemented complete error handling, validation, and logging for the `POST /invites` router and `createClinicianInvite` service in PI DAS using `consent-api` as an example.

### PI DAS
- Added `express-request-validation` from shared packages to use `withRequestValidation` to validate the endpoint's request body
- Updated the endpoint's HTTP Responses to use the appropriate `ErrorResponse`
- Added a 400 RequestValidationError and 409 ConflictError to the endpoint responses, updated the JSDocs
- Updated the return type of the `createClinicianInvite` service to `Result` so it returns appropriate `Success` and `Failure` responses to the router
- Updated the Prisma create request so it returns a failure response if an error occurred
>**Note:** The parameter type for `createClinicianInvite` was changed to the Zod schema `PIClinicianInvite` so it no longer uses the `clinicianId` field if a manually generated ID is being passed to it

### Special Instructions
Before running these changes, you will need to add `express-request-validation` to PI DAS:
```
pnpm i
```

## PR Readiness Checklist

- [x] Connected ticket to PR

### DIFF
--- a/apps/pi-das/src/service/create.ts
+++ b/apps/pi-das/src/service/create.ts
@@ -1,4 +1,11 @@
+import { PIClinicianInviteRequest } from 'types/entities';
+import { Result, failure, success } from 'types/httpResponses';
+
 import prisma, { Participant, Province, ClinicianInvite } from '../prismaClient.js';
+import { PrismaClientKnownRequestError } from '../generated/client/runtime/library.js';
+import serviceLogger from '../logger.js';
+
+const logger = serviceLogger.forModule('PrismaClient');
 
 export const createParticipant = async ({
 	inviteId,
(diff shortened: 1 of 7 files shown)


mistryrn (dismissed):
Looks good to me! ✅

apps/pi-das/src/service/create.ts
```diff
@@ -65,42 +72,37 @@ export const createParticipant = async ({
...
+type CreateInviteFailureStatus = 'SYSTEM_ERROR' | 'INVITE_EXISTS';
+/**
+ * Creates a ClinicianInvite entry in the PI DB
+ * @param inviteRequest Clinician Invite data
+ * @returns ClinicianInvite object from PI DB
+ */
+export const createClinicianInvite = async (
+	inviteRequest: PIClinicianInviteRequest,
+): Promise<Result<ClinicianInvite, CreateInviteFailureStatus>> => {
+	return await prisma.clinicianInvite
```
anncatton, line 84: i think this `await` is not required since we have a `then()` and we're returning a Promise?

apps/pi-das/src/service/create.ts
```diff
@@ -65,42 +72,37 @@ export const createParticipant = async ({
...
+						'INVITE_EXISTS',
+						`An invite already exists with that '${error.meta?.target ?? 'data'}'`,
+					);
+				}
+				logger.error(
+					'POST /invites',
+					'Unexpected error handling create invite request.',
+					error.message,
+				);
+				return failure('SYSTEM_ERROR', 'An unexpected error occurred.');
```
anncatton, line 105: i'm thinking here it might be good to instead split up this if clause:
``` 
if (error instance of PrismaClientKnownRequestError) {
          if (error.code === 'P2002') { ...do conflict handling }
          // an unexpected prismaClient error
         logger.error('POST /invites', error.code)
          return failure('SYSTEM_ERROR', 'An unexpected error occurred in the PrismaClient.', error.code); 
}
```
and then add a catch statement that returns the generic systemError message in the case where something just blows up
```
catch (err) { return failure('SYSTEM_ERROR', 'An unexpected error occurred.'); }
```

anncatton (commented):
Just some comments about the error handling - let me know what you think, in case i'm missing something about that prisma client call here!

apps/pi-das/src/service/create.ts
alyssadsouza: Makes sense! Also, it seems that passing the second argument to `then()` (the callback to execute when the promise is rejected) behaves the same way as having a normal `catch` block would, so I moved that error handling callback to the `catch` for better readability.

Issue #255, opened 2023-11-16
Build Data Mapper Endpoint --> Consent Wizard - GET - Step 1


mistryrn:
Build corresponding `GET` endpoint in the `data-mapper` to retrieve the data for Step 1 (Informed Consent) and serve it back to the `consent-api` as a Zod object.

## DAS Breakdown
- Consent DAS
  - [x] ParticipantResponse
  - [x] ConsentQuestion

## Data Mapper Breakdown
- [x] Add `getInformedConsentResponses` service to `services/search.ts`

## Order of Operations
1. Get consent questions for the Informed Consent category from the Consent DAS
2. Get participant responses for each of the retrieved consent questions from the Consent DAS
3. Get most recent response per consent question
4. Serve in response

## Success Response
`200` containing the Informed Consent object

## Other Considerations
Highlight areas that may need the following:
- Logging
- Error Handling
- Validation

b-f-chan:
Done, closing

PR #316, opened 2023-12-13
✨ #255: Data Mapper GET route for Informed Consent responses

alyssadsouza:
## Description of Changes

Implements the `GET /wizard/steps/informed-consent/:participantId` route and the services it calls in Data Mapper, which retrieves the most recent responses to each question in the Informed Consent step (step 1) of Consent Wizard. Includes error handling, validation, and logging.

### Data Mapper
**Router**
- Added the `wizard` router and nested the `steps` and `informed-consent` router
- Added `GET` endpoint to the `informed-consent` router to retrieve the responses, with error handling for:
  - `400: RequestValidationError` (invalid nano ID for participant ID)
  - `404: NotFoundError` (participant does not exist) 
  - `500: ServerError` (unexpected error)

**Services**
- Added the `getConsentQuestionsByCategory` service which makes a request to Consent DAS to fetch all consent questions from a category
- Added the `getParticipantResponsesByQuestionId` service which makes a request to Consent DAS to fetch all participant responses for a given question in most to least recent order
- Added  the`getParticipantResponsesByCategory` service which calls `getConsentQuestionsByCategory` to first retrieve all questions for the specified category, and then `getParticipantResponsesByQuestionId` for each question to fetch and return the most recent participant responses for the category
- Added the `getInformedConsentResponses` service which calls `getParticipantResponsesByCategory` with the Informed Consent category to return the most recent Informed Consent participant responses

### Types
- Removed the `InformedConsentBase` schema and updated `InformedConsentResponse` to have its responses be `literal(true).optional()` and `InformedConsentRequest` to have its responses be `literal(true)`
- Added an `isSuccess` method to the `Results` types to make it more convenient (and TS inferable) to check for `Failure` types; updated the error handling docs

## PR Readiness Checklist

- [x] Connected ticket to PR


packages/types/src/entities/InformedConsent.ts
```diff
@@ -24,7 +24,7 @@ import type { SchemaObject } from 'openapi3-ts/oas31';
 import { ConsentQuestionId } from './ConsentQuestion.js';
 
 export const InformedConsentBase = z.object({
-	[ConsentQuestionId.enum.INFORMED_CONSENT__READ_AND_UNDERSTAND]: z.boolean(),
+	[ConsentQuestionId.enum.INFORMED_CONSENT__READ_AND_UNDERSTAND]: z.boolean().optional(), // TODO: verify if this can be optional
```
alyssadsouza, line 27: If they haven't responded to a question, I'm assuming `{ questionId: undefined }` (which becomes `{}`) is alright for the response, unless we would want the question ID to still be present and instead flag a non response with something else? But I can't think of a reason why we would need to have all the question IDs present.

alyssadsouza:
@mistryrn it would return an empty array, actually thanks for pointing that out because I'll need to add some handling here when that's the case https://github.com/OHCRN/platform/pull/316/files#diff-0ef30b90fae0aff4a3f6987bb50fa39b6d3bd5d3353cf9e074677eeb57d94d36R121 - so should it just be `[consentQuestionId] = undefined`?

apps/data-mapper/src/services/das/consent.ts
```diff
@@ -104,3 +111,99 @@ export const createInviteConsentData = async (
...
+ * @param category Consent category i.e. step name in Consent Wizard
+ * @returns Consent questions for the category
+ */
+export const getConsentQuestionsByCategory = async (
+	category: ConsentCategory,
+): Promise<Result<ConsentQuestionArray, GetResponsesFailureStatus>> => {
+	try {
+		const { consentDasUrl } = getAppConfig();
+		const { data } = await axiosClient.get(
+			urlJoin(consentDasUrl, 'consent-questions', `?category=${category}`),
```
anncatton, line 126: a nicer way to do this would be to use the `URL` constructor: https://developer.mozilla.org/en-US/docs/Web/API/URL
then you can add in params with `append`, like: 
```
const foo =  new URL(urlJoin(consentDasUrl, 'consent-questions'))
foo.searchParams.append('category', category)
```

apps/data-mapper/src/routers/steps/informedConsent.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+router.get('/:participantId', async (req, res) => {
+	try {
+		const { participantId } = req.params;
+
+		const requestParticipantId = NanoId.safeParse(participantId);
+
+		if (!requestParticipantId.success) {
+			logger.error(
+				'GET /wizard/steps/informed-consent/:participantId',
```
anncatton, line 74: just thinking a minor enhancement we can add is, set the router path at the top of the file to use in these logs, i.e. `const ROUTER_PATH = '/wizard/steps/informed-consent'`

apps/data-mapper/src/services/search.ts
```diff
@@ -62,7 +76,97 @@ export const getParticipant = async (participantId: string): Promise<any> => {
...
+		const participantResponses: ParticipantResponsesByCategory = {};
+
+		for (const consentQuestion of consentQuestions.data) {
+			const responsesResult = await getParticipantResponsesByQuestionId({
+				participantId,
+				consentQuestionId: consentQuestion.id,
+			});
+
+			if (responsesResult.status !== 'SUCCESS') {
+				return responsesResult;
```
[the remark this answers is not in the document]
alyssadsouza: Good question, since it's in a normal for loop my understanding is that it exits the whole loop i.e. returns the function value (TS also agrees with me because if I try and return an arbitrary value that doesn't match the return type of the function it complains). As to whether we should... I assumed we should because the entire API response is malformed at that point – it won't pass validation if it can't find all the questions and corresponding responses, so ending the function there right away and returning the `failure` that resulted from the response it can't find makes sense to me. I'm open to suggestions!

packages/types/src/entities/InformedConsent.ts
alyssadsouza: discussed on slack and decided on `undefined` for no responses

packages/types/src/httpResponses/Result.ts
```diff
@@ -38,6 +38,18 @@ export type Result<T, FailureStatus extends string, FailureData = void> =
    Convenience Methods 
  * ******************* */
 
+/**
+ * Determines if the Result is a Success type by its status
+ * and returns the type predicate so TS can infer the Result as a Success
+ * @param result
+ * @returns Whether the Result was a Success or not
+ */
+export function isSuccess<T, FailureStatus extends string, FailureData>(
```
mistryrn, line 47: This is super handy!

mistryrn (approved):
Thanks for bearing with us and getting those changes in! This looks good to me now, maybe just give @anncatton the chance to re-review in the morning before merging :)

packages/types/src/httpResponses/Result.ts
```diff
@@ -38,6 +38,18 @@ export type Result<T, FailureStatus extends string, FailureData = void> =
    Convenience Methods 
  * ******************* */
 
+/**
+ * Determines if the Result is a Success type by its status
+ * and returns the type predicate so TS can infer the Result as a Success
+ * @param result
+ * @returns Whether the Result was a Success or not
```
anncatton, line 45: `@returns {boolean} Whether the Result was a Success or not`

anncatton (approved):
just 2 very smol comments!

anncatton (approved):

Issue #170, opened 2023-10-24
Build Consent API Endpoint --> Consent Wizard - GET - Step 1


b-f-chan:
Build consent API endpoint: `GET /wizard/steps/informed-consent`

**Applicable Page:** Consent Wizard - Informed Consent (step 1)
**Functionality:** Get consent questions for Informed Consent wizard step

**Expected Outcome(s):**

- [x] Swagger doc for the endpoint is functional and testable
- [x] Unit testing implemented for this endpoint

---

Update 10/23/2023 from Sam:

- Endpoint mapped out here https://github.com/OHCRN/platform/pull/202

b-f-chan:
Merged, closing

PR #318, opened 2023-12-14
✨ #170: Consent API GET route for Informed Consent responses

alyssadsouza:
## Description of Changes

Implements the `GET /wizard/steps/informed-consent/` route and the service it calls `getInformedConsentResponses`, which makes a request to Data Mapper to retrieve the most recent responses to each question in the Informed Consent step (step 1) of Consent Wizard. Includes error handling, validation, and logging. **This endpoint is incomplete until authentication is implemented, because the `participantId` that will be used to retrieve the responses for will be accessed from the session token.**

### Consent API
- Added `GET` endpoint to the `informed-consent` router to retrieve the responses, with error handling for:
  - `400: RequestValidationError` (invalid nano ID for participant ID)
  - `500: ServerError` (unexpected error)
- Added the `getInformedConsentResponses` service which makes a request to Data Mapper's `GET /wizard/steps/informed-consent/:participantId` endpoint
- Added unit testing for the endpoint in `tests/routers/steps/informed-consent.spec.ts`

### Other Changes
- Added a README in the Consent API `tests` dir to document how we're unit testing our routers

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-api/src/services/search.ts
```diff
@@ -30,3 +42,42 @@ export const getLatestParticipantResponseByParticipantIdAndQuestionId = async (
 	// TODO: fetch from data-mapper, then return the result
 	return { participantId, consentQuestionId };
 };
+
+export type GetResponsesFailureStatus = 'SYSTEM_ERROR';
+
+/**
+ * Retrieves most recent responses for each INFORMED_CONSENT question
```
anncatton, line 49: Love the attention to detail! But i think to keep the jsdocs clear let's just document what this function is doing, not what happens further down the line

apps/consent-api/tests/README.md
```diff
@@ -0,0 +1,46 @@
...
+console.log(response.status); // 200
+```
+
+## Mocking services
+
+An endpoint will likely make a call to a service function (e.g. [`services/create.ts`](../src/services/create.ts)). Mock service functions as needed using `vi.mock()` so you don't need to make an API call to other services like Data Mapper or the DAS (which will be running independent of Consent API).
+
+### Using variables in `vi.mock()`
+
+`vi.mock()` gets "hoisted" to the top of the file meaning it executes first, so you'll need to declare any variables used in `vi.mock()` inside of a `vi.hoisted()` function call before declaring your mocks. You can read more about it in the [docs](https://vitest.dev/api/vi#vi-mock). As an example, it might look like this:
```
anncatton, line 23: 🙏

apps/consent-api/src/services/search.ts
```diff
@@ -30,3 +42,42 @@ export const getLatestParticipantResponseByParticipantIdAndQuestionId = async (
...
+ * by making request to Data Mapper, which fetches all questions for the category
+ * and then fetches the corresponding response per question
+ * @param participantId ID of participant to retrieve responses for
+ * @returns
```
anncatton, line 53: as discussed over zoom, let's add the return type here even tho we still need to figure out why intellisense doesn't display it as expected

apps/consent-api/src/services/search.ts
```diff
@@ -30,3 +42,42 @@ export const getLatestParticipantResponseByParticipantIdAndQuestionId = async (
...
+ */
+export const getInformedConsentResponses = async (
+	participantId: string,
+): Promise<Result<InformedConsentResponse, GetResponsesFailureStatus>> => {
+	try {
+		const { dataMapperUrl } = getAppConfig();
+		const { data } = await axiosClient.get(
+			urlJoin(dataMapperUrl, 'wizard', 'steps', 'informed-consent', participantId),
```
anncatton, line 61: i see this endpoint is used in the test file, maybe we should have the path value (or parts of it) saved somewhere as a constant?

anncatton (commented):
super minor comments/suggestions but otherwise looks good

apps/consent-api/src/services/search.ts
alyssadsouza: the one thing I'll say for that though is the one in the test file is pointing to the route in Consent API whereas this is actually pointing to the Data Mapper route

anncatton: gotcha gotcha let's leave it then

mistryrn (approved):

Issue #233, opened 2023-11-06
Style Clinician Invite Form


b-f-chan:
**Task:**

- [ ] Setup reusable components with styling on the Clinician Invite form:
    - [ ] Text input
    - [ ] Checkbox
    - [ ] Dropdown 
    - [ ] Tooltips

**Pre-Requisite:**

- Assuming the form fields and logic exist already in the Sample Page ticket (#186)

**Expected Outcome(s):**

- [x] Have a clinician invite page where the user can fill out the form
- [x] Data should be validated on the form
- [ ] When user clicks submit, it will log results in the browser console output

b-f-chan:
To close this one, need to merge #152, fixing the 1024 issue

b-f-chan:
@samrichca can this be closed?

PR #323, opened 2023-12-18
💄 #233: Style invite form

samrichca:
### Consent-UI

- update non-form components
  - update Button component to work as a form submit button
  - allow Modal component to have scrolling content
  - fix issue in SideImageLayout where screens 1024px wide had no visible navigation
- update dictionaries
  - translate consent groups FR file
  - compose invite dictionaries from generic dictionaries
  - rename `invite` dictionary to `inviteFormPage` for clarity
- add form styling
  - make RequiredAsterisk bold
  - update styling for checkbox, text, and select inputs
    - add disabled & error states
    - add narrow desktop option, to retain tablet-size styling at desktop size
    - add optional callout to form fields, which is shown on focus but remains in the markup at all times for screenreaders
    - add optional info button to field label component
- build clinician invite form 
  - [use react `useId` to create unique IDs that persist between renders](https://react.dev/reference/react/useId#useid)
  - move single-use types to their respective components
  - add modal for consent group descriptions
- update readme 

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/components/common/Button/index.tsx
```diff
@@ -25,8 +25,9 @@ import { ButtonProps as BaseProps } from './types';
 import styles from './Button.module.scss';
 
 interface ButtonProps extends BaseProps {
-	onClick: (e: React.SyntheticEvent<HTMLElement>) => any;
+	onClick?: (e: React.SyntheticEvent<HTMLElement>) => any;
```
samrichca, line 28: submit buttons in our forms can't have onClick events & must have `type="submit"`

apps/consent-ui/src/components/layouts/SideImageLayout/SideImageLayout.module.scss
```diff
@@ -88,18 +92,19 @@
...
-	min-height: 6rem;
-}
-
-.desktopHeader,
-.desktopNav {
-	@media (max-width: theme('screens.md')) {
-		display: none;
+	min-height: 4rem;
+	display: none;
+	@media (min-width: theme('screens.md')) {
```
samrichca, line 101: original media queries for this component created an issue where there was no navigation at all at exactly 1024px screen width

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -97,6 +109,7 @@ const ClinicianInviteFormComponent = ({
 		const recaptchaToken = getRecaptchaToken();
 
 		if (recaptchaToken) {
+			console.log('form data', data);
```
samrichca, line 112: this console log is part of the ticket requirements because the API call doesn't work yet

ciaranschutte (commented):
Copyrights can be 2024 now instead of 2023

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -37,8 +38,17 @@ import Form from 'src/components/common/Form';
...
+import Button from 'src/components/common/Button';
+import layoutStyles from 'src/components/layouts/SideImageLayout/SideImageLayout.module.scss';
+import { useModal } from 'src/components/common/Modal';
+import ConsentGroupModal from 'src/components/views/Invite/ConsentGroupModal';
+import { ValidLanguage } from 'src/i18n';
 
 import { ConsentGroupOption } from './types';
+import styles from './ClinicianInviteForm.module.scss';
+
+Object.assign(styles, layoutStyles);
```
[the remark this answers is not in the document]
samrichca: nah, this is the documented way to use multiple CSS modules: https://github.com/css-modules/css-modules/blob/master/docs/import-multiple-css-modules.md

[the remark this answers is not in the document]
justincorrigible: does this mutation possibly affect the imported `styles` in other contexts?

even though documentation may use that pattern, I'm inclined to suggest we'd stick to immutable patterns for the sake of consistency... but I'm not certain whether that's just my Functional Programming stickler side speaking.

```suggestion
const inviteFormStylesOrWhatever = Object.assign({}, styles, layoutStyles);
```

apps/consent-ui/src/components/common/Form/FormSection/FormSection.module.scss
```diff
@@ -0,0 +1,16 @@
+.formSection {
+	width: calc(100% + 2rem);
```
mistryrn, line 2: Just out of curiosity, is there a reason for adding the extra width and negative margin here? Versus just specifying the top and bottom padding only:
```css
padding: 1rem 0;
```

I may have missed something in the designs, or maybe there's a specific responsive layout where the negative margin is necessary here so please let me know

samrichca: it's to line up the grey sections with the rest of the page 
image

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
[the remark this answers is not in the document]
samrichca: `Object.assign(styles, layoutStyles)` mutates the imported styles object in this file and doesn't affect global scope

updated the code to remove the mutation

apps/consent-ui/src/components/common/Form/FormSection/FormSection.module.scss
mistryrn: Ah, that makes sense. Thanks!

apps/consent-ui/src/components/common/Form/fieldsets/SelectFieldSet/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+	const idPrefix = useId();
+	const fieldId = `${idPrefix}-${name}`;
+	const calloutId = `${idPrefix}-callout`;
+
+	return (
+		<FieldSet
+			calloutContent={calloutContent}
+			calloutId={calloutId}
+			calloutVisible={calloutVisible}
+			className={clsx(styles.selectFieldSet, className)}
```
mistryrn, line 65: Looks like we're not passing the `error` prop down here, so validation errors never render.

```suggestion
			className={clsx(styles.selectFieldSet, className)}
			error={error}
```

apps/consent-ui/src/components/common/Form/fieldsets/TextFieldSet/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const idPrefix = useId();
+	const fieldId = `${idPrefix}-${name}`;
+	const calloutId = `${idPrefix}-callout`;
+
+	return (
+		<FieldSet
+			calloutContent={calloutContent}
+			calloutId={calloutId}
+			calloutVisible={calloutVisible}
+			className={className}
```
mistryrn, line 60: Same as the `SelectFieldSet`, looks like we're not passing the `error` prop down into the `FieldSet` here:
```suggestion
			className={className}
			error={error}
```

apps/consent-ui/src/components/common/Modal/index.tsx
```diff
@@ -81,35 +81,35 @@ const ModalProvider = ({ children }: { children: ReactNode }) => {
 	return (
 		<ModalContext.Provider value={value}>
 			{isOpen && (
-				<div className={styles.modal}>
-					<div ref={ref}>
```
mistryrn, line 85: Looks like removing this extra div and moving the ref onto the wrapper breaks the functionality of the `useDetectClickOutside` library that Alyssa added.

I don't know why, but I've seen it with other 3rd party libraries for this kind of thing before (I think there was something in the old platform-ui that behaved similary – I removed what I thought was a redundant/pointless div and all of a sudden the functionality was borked 😅)

If we restore that extra div and move the ref back onto it instead of the wrapper, clicking outside closes the modal again. But if you can figure out why/find a better solution that would be acceptable too!

docs/forms.md
mistryrn, line 1: Thanks for updating these docs!

mistryrn (changes requested):
Really great work! 😃 Found a few small things during my testing, as noted in the comments below, but the UI is really starting to come together! Looks great!!

Aside from the comments below, I do think we should rename `Callout` to `Tooltip` (and the related functions, styles, and translations). Sorry, I know it's a pain! 😬 But I think this component better matches the definition of a Tooltip than a Callout: https://medium.com/@adamshriki/choosing-the-right-ui-component-for-communication-with-your-users-523499c39490

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
```diff
@@ -19,39 +19,72 @@
...
+			render={({ field: { onChange, value } }) => (
+				<Select
+					aria-required={required}
+					className={clsx('react-select-container', hasError && 'react-select__has-error')}
+					classNamePrefix="react-select"
+					// react-select doesn't work with hashed CSS classnames from CSS modules.
+					// className & classNamePrefix need to be strings, not hashed classNames.
+					// https://github.com/JedWatson/react-select/issues/4525
+					id={`select-${id}`}
+					inputId={`input-${id}`}
```
mistryrn, line 63: It looks like now that we're using an id generated from `useId` here instead of the `id` prop passed in from `SelectFieldSet` (https://github.com/OHCRN/platform/blob/4fc1799f304fa4d2dafc92591f5b49a7f9198517/apps/consent-ui/src/components/common/Form/fieldsets/SelectFieldSet/index.tsx#L77), the `for` prop on the label that the `FieldSet` creates no longer points at the generated select input.

Screenshot 2024-01-10 at 11 28 44 AM

We're getting the same a11y error we were getting before (https://github.com/OHCRN/platform/pull/235#discussion_r1417987720) on the select inputs, `Form elements must have labels`.

Screenshot 2024-01-10 at 11 31 29 AM

mistryrn (commented):
Thanks for getting those changes in! Just one a11y regression to address in the `SelectInput`, and a few minor case changes needed after the renaming from callout to tooltip 🙂

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
samrichca: good catch, updated

apps/consent-ui/src/components/common/Form/fieldsets/Tooltip/useTooltip.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+'use client';
+
+import { useState } from 'react';
+
+const useTooltip = () => {
+	const [tooltipVisible, settooltipVisible] = useState<boolean>(false);
```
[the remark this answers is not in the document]
samrichca: thanks, fixed

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
```diff
@@ -19,39 +19,72 @@
...
+					// react-select doesn't work with hashed CSS classnames from CSS modules.
+					// className & classNamePrefix need to be strings, not hashed classNames.
+					// https://github.com/JedWatson/react-select/issues/4525
+					id={`select-${id}`}
+					inputId={id} // must match label htmlFor
+					instanceId={`instance-${id}`}
+					isDisabled={disabled}
+					name={name}
+					onBlur={onBlur}
+					onChange={(val: SelectOnChangeArg<V>) => {
```
anncatton, line 67: i think the typing on the `val` arg here is unnecessary, looks like it already shows as `SingleValue<string | FormSelectOption<V>>` in Intellisense

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
```diff
@@ -19,39 +19,72 @@
...
+						// in react-select the value can be a string or object.
+						// in our implementation it must be {label, value},
+						// with the label being translated.
+						let onChangeParam = '';
+						if (typeof val === 'string') {
```
anncatton, line 72: with the type of `val` being `SingleValue<string | FormSelectOption<V>>`, i'm not sure this `if` clause would ever be true? i'm assuming the empty string case is for when and option hasn't yet been selected (?)

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
```diff
@@ -19,39 +19,72 @@
...
-import { FieldValues } from 'react-hook-form';
-import Select from 'react-select';
+import clsx from 'clsx';
+import { Controller, FieldValues, useFormContext } from 'react-hook-form';
+import Select, { SingleValue } from 'react-select';
 
-import { FormSelectInputProps, FormSelectOnChangeArg } from '../../types';
+import { FormInputProps, FormSelectOption } from 'src/components/common/Form/types';
+
+type SelectOnChangeArg<V extends string> = SingleValue<string | FormSelectOption<V>>;
```
anncatton, line 28: may not need this type, see comment below

apps/consent-ui/src/components/common/Form/fieldsets/inputs/SelectInput.tsx
samrichca: correct about the empty string case, I added a comment to clarify this

anncatton: is the `typeof val === 'string'` relevant here tho? it looks like the return should either be `onChange(val.value)` or `onChange('')`. Or perhaps i'm missing something..?

samrichca: if I don't handle the string case explicitly, I get a TS issue

```ts
let onChangeParam = val;
						if (val?.value !== undefined) {
							onChangeParam = val.value;
						}
```

`Property 'value' does not exist on type 'string | FormSelectOption<V>'. Did you mean 'valueOf'?
  Property 'value' does not exist on type 'string'.`

anncatton: hmm yeah i just tried locally and see what you mean...i guess there is something in that `SingleValue` type i'm not getting. 🤷‍♀️

anncatton (commented):
lgtm, but i'll let @mistryrn do final approval

mistryrn (approved):
Looks good! 🚀

Issue #313, opened 2023-12-12
Add function to translate entire namespace


samrichca:
Add a function that translates an entire namespace, with parameters.

## Exit Criteria

A function similar to `translate()` that

- returns an entire namespace's translations, in an object
- accepts parameters to replace parts of the translated string(s)

e.g.

```
translateAll('myTranslationNamespace', { translationKey: { year: "2023" }});

/// result

{
  translationKey: "This text was translated in the year 2023.",
}
```

b-f-chan:
Do during translation sprint before launch

b-f-chan:
Merged, closing

PR #356, opened 2024-01-10
🌐 #313: add function to translate namespace

samrichca:
### Consent UI

- add a function that translates an entire namespace
- refactor to use translateNamespace in places where the same dictionary is translated multiple times

## PR Readiness Checklist

- [x] Connected ticket to PR


samrichca:
@anncatton good point about enforcing server only for translations, we should get on that

I'll update the places where I'm not fetching the entire dictionary.

docs/translations.md
```diff
@@ -55,7 +55,7 @@ const frenchCommonDictionary = { ... } satisfies CommonDictionary;
```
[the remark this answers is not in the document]
samrichca: updated readme and added jsdoc comments

docs/translations.md
```diff
@@ -74,22 +75,46 @@ const translatedString = translate('footer', 'about');
...
 ```typescript
-const translate = getTranslation('fr');
+const { translate } = getTranslation('fr');
 const translatedString = translate('footer', 'copyright', { year: '2023' });
 // => '© 2023 Ontario Hereditary Cancer Research Network. Tous droits réservés.'
 ```
 
+#### Translating a Namespace
+
+Translate an entire namespace at once when you want to pass down many translations from a server component to a client component, e.g. for forms.
```
anncatton, line 85: this is great, thanks for adding. One smol request, can you add a note about the "why" here - we are keeping the translations on the server side, so we can't access the functions in these client components. `translateNamespace` is a convenience function to retrieve all necessary translations in a server parent, and pass them to a child client component.

apps/consent-ui/src/i18n/utils/getTranslation.ts
```diff
@@ -56,7 +56,18 @@ export const replaceParams = (
 // TODO: is there a way to enforce this function for server side use only?
 export const getTranslation: GetTranslation = (language) => {
 	const dictionary = dictionaries[language];
-	return (namespace, key, params) => {
+
+	/**
```
anncatton, line 60: adding jsdocs is great. it looks though like Intellisense is not able to pick up the comments when they are nested inside a function like this. Poking around I didn't see any satisfactory solution...i think just as a little helper here, we can add a simple jsdoc to `getTranslation`, maybe something like:
```
/**
 * @param language ValidLanguage
 * @exports {translate, translateNamespace}
 * /
```

anncatton (commented):
good stuff. just a couple minor comments and then should be good to go!

anncatton:
> @anncatton good point about enforcing server only for translations, we should get on that
> 
> I'll update the places where I'm not fetching the entire dictionary.

Forgot to add in my other review, i'll make a separate ticket for the server-only setup and we can discuss at planning next week

anncatton:
Created a ticket for the server-only task https://github.com/OHCRN/platform/issues/422

anncatton (approved):

PR #386, opened 2024-01-22
✨♻️🏷️  254 - data mapper create participant

anncatton:
## Description of Changes

Implements `POST` `/participants` endpoint in Data Mapper for participant creation, as part of the registration flow.
Adds schema types to support these changes.

*Note: did not complete manual testing of the `POST` endpoint because there are changes required in both the PI and Consent DASes to support this flow. These will be implemented in subsequent tickets.

### Types

- moves guardianInfo check to utils file in `common` folder
- adds participant contact info (email, phoneNumber) check function. TBD whether these fields will be made optional when a guardian is present (#388 )
- adds modular schemas for Participant, in its own entity file
- updates Participant fields to match what is needed in registration flow
- renames `ParticipantIdentification` to `ParticipantIdentity`, and moves to its own entity file
- creates a separate `Guardian` entity file for code cleanup. This file includes a plain object for converting the optional fields being returned from the db
- adds request and response types to `services/data-mapper` for `CreateParticipant`, `ConsentCreateParticipant` (fields for consent-das) and `PICreateParticipant` (fields for pi-das)
- refactors `PIClinicianInvite` (`pi-das/responses`) and `ClinicianInvite` entity to use new Participant + Guardian sub schemas
- updates and cleans up `ClinicianInvite` and `ParticipantIdentity` tests
- changes regex imports to come directly from the `regexes.ts` file to fix an import resolution error
- adds `TODO`s for open questions being tracked in #388  

### Data Mapper

- moves existing DAS service calls to their own files
- updates `/participants` `POST` endpoint with request validation, error handling and jsdocs
- adds CreateParticipant schemas to swagger router
- updates router and service function parameters
- implements `createParticipant` service function that creates the participant in both PI and Consent DASes, and rolls back the changes if an error occurs.
- implements `deletePiParticipant` to support rollback in `createParticipant`
- implements separate service functions for creating participants in the PI and Consent DASes, with validation and error handling
- moves existing PHI and Keys DAS functions to their own `/services/das/` files. Functionality will be properly implemented in future tasks
- adds `TODO`s for open questions being tracked in #388 

## PR Readiness Checklist

- [x] Connected ticket to PR
- [ ] Manual testing completed


apps/data-mapper/src/routers/participants.ts
```diff
@@ -46,19 +49,59 @@ router.get('/:participantId', async (req, res) => {
...
+			switch (participant.status) {
+				case 'SUCCESS': {
+					return res.status(201).json(participant.data);
+				}
+				case 'PARTICIPANT_EXISTS': {
+					return res.status(409).json(ConflictErrorResponse(participant.message));
+				}
+				case 'SYSTEM_ERROR': {
+					return res.status(500).json(ServerErrorResponse(participant.message));
+				}
```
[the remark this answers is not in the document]
joneubank: Looks like we're missing a case for `InvalidRequest` which is a valid failure state for `CreateParticipantFailureStatus`. I suspect the solution is to remove `InvalidRequest` from that state union since it is never returned by `createParticipant`

apps/data-mapper/src/services/create.ts
```diff
@@ -1,112 +1,123 @@
...
-// TODO: add Type instead of any
-const createParticipantOhipKey = async (participantId: string): Promise<any> => {
-	const { keysDasUrl } = getAppConfig();
-	// TODO: add error handling
-	const result = await axiosClient.post(urlJoin(keysDasUrl, 'ohip-keys'), {
-		participantId,
-	});
-	return result.data.ohipKey;
-};
+export type CreateParticipantFailureStatus = SystemError | InvalidRequest | 'PARTICIPANT_EXISTS';
```
joneubank, line 22: As mentioned in other comment, it looks like `InvalidReqeuest` is not needed here.

More importantly, these appear to be the failure status for the PI client request to create participant, so this type should be declared alongside that function. It can be reused here if desired. The problem right now is that `services/create.ts` is importing a function from `services/das/pi.ts`, while this type is being imported the otherway. This is a cycle in our imports. Cycles in imports is a "code smell" giving an indication that there is a logic problem in our code structure.

apps/data-mapper/src/services/create.ts
```diff
@@ -1,112 +1,123 @@
...
+			currentLifecycleState,
+			emailVerified,
+		});
+
+		if (participantConsentData.status !== 'SUCCESS') {
+			// Unable to create participant in Consent DAS, rollback participant already created in PI-DAS
+			const deletePiParticipant = await deletePIParticipant(participantId);
+			if (deletePiParticipant.status !== 'SUCCESS') {
+				logger.error('Error deleting existing PI participant:', deletePiParticipant.message);
+				return failure('SYSTEM_ERROR', 'An unexpected error occurred.');
```
joneubank, line 103: We could probably improve this logging. This puts our data in an unexpected state, with the PI participant created but no consent data. It is probably recoverable but this state should be captured better in our log. Consider including participant ID so we know which ID to take action on if we are searching for this in our logs.

joneubank: As a next step after this PR... if our data is in this state, can the user resubmit their data successfully or will be wind up conflicting with "participant exists" type errors? Is action needed?

apps/data-mapper/src/services/das/consent.ts
```diff
@@ -18,19 +18,21 @@
...
 } from 'types/dataMapper';
+import { ConsentCategory, ConsentQuestionArray, ConsentQuestionId } from 'types/entities';
+import { Result, failure, success } from 'types/httpResponses';
 import urlJoin from 'url-join';
 
 import { getAppConfig } from '../../config.js';
 import serviceLogger from '../../logger.js';
 import axiosClient from '../axiosClient.js';
-import { CreateInviteFailureStatus } from '../create.js';
+import { CreateInviteFailureStatus, CreateParticipantFailureStatus } from '../create.js';
```
joneubank, line 35: As mentioned in comment above, the failure status type should probably be declared in this module along side `createParticipantConsentData`.

apps/data-mapper/src/services/das/keys.ts
```diff
@@ -16,3 +16,19 @@
...
  */
+
+import urlJoin from 'url-join';
+
+import { getAppConfig } from '../../config.js';
+import axiosClient from '../axiosClient.js';
+
+// TODO: add proper JSDoc comments
+// TODO: add Type instead of any
+export const createParticipantOhipKey = async (participantId: string): Promise<any> => {
```
joneubank, line 27: You're not asking me to approve a PR with a `Promise<any>` return type, are you? :P

joneubank: lol TODO

apps/data-mapper/src/services/das/pi.ts
```diff
@@ -136,3 +141,102 @@ export const deleteInvitePiData = async (
...
+			participantOhipLastName,
+			dateOfBirth,
+			participantPhoneNumber,
+			participantPreferredName,
+			participantEmailAddress,
+			guardianEmailAddress,
+			guardianPhoneNumber,
+			guardianRelationship,
+			keycloakId,
+		});
```
joneubank, line 179: As far as i can tell, you never use these destructured properties individually. Why not just use `req` as an object instead of listing all the members?

apps/data-mapper/src/services/das/pi.ts
```diff
@@ -136,3 +141,102 @@ export const deleteInvitePiData = async (
...
+			const participantId = data.participant.id;
+			if (participantId) {
+				// delete participant from PI DB if response parsing fails
+				logger.info('Deleting invalid participant');
+				const deletePiParticipant = await deletePIParticipant(participantId);
+				if (deletePiParticipant.status !== 'SUCCESS') {
+					logger.error('Error deleting existing PI participant:', deletePiParticipant.message);
+					return failure('SYSTEM_ERROR', 'An unexpected error occurred.');
+				}
+			}
```
joneubank, line 196: This block surprises me. Are we preparing for a case where the request was successful but the response was invalid, so we create and then delete the participant? I actually think this might not be necessary. Let the PI service decide if creating the participant is correct. If the messaging fails to communicate the info about the participant then we handle this as a system error in everything downstream. Consider this similar to failing to fetch data, you don't request to delete data if the response is malformed.

apps/data-mapper/src/services/das/pi.ts
```diff
@@ -136,3 +141,102 @@ export const deleteInvitePiData = async (
...
+	} catch (error) {
+		if (error instanceof AxiosError && error.response) {
+			const { data, status } = error.response;
+			logger.error('AxiosError handling delete participant request', data);
+
+			if (status === 404) {
+				return failure('PARTICIPANT_DOES_NOT_EXIST', data.message);
+			}
+
+			return failure('SYSTEM_ERROR', data.message);
```
joneubank, line 237: 👍

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+	ADULT_CONSENT,
+	YOUNG_ADULT_CONSENT,
+} = ConsentGroup.enum;
+
+const requiresGuardianInformation = (consentGroup: ConsentGroup) => {
+	return (
+		consentGroup === GUARDIAN_CONSENT_OF_MINOR ||
+		consentGroup === GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT ||
+		consentGroup === ADULT_CONSENT_SUBSTITUTE_DECISION_MAKER
+	);
```
joneubank, line 36: Consider the following pattern to make this easier to extend in the future and prevent boolean logic typos.

```ts
const consentOptions: ConsentGroup[] = [GUARDIAN_CONSENT_OF_MINOR, GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT, ADULT_CONSENT_SUBSTITUTE_DECISION_MAKER];

consentOptions.includes(consentGroup);
```

joneubank: Note that the type of `consentOptions` needs to be `(typeof consentGroup)[]`, so in this case `ConsentGroup[]` so that the TS won't have issues with `consentOptions.includes(consentGroup)`.

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const requiresParticipantContactInfo = (consentGroup: ConsentGroup) => {
+	return consentGroup === ADULT_CONSENT || consentGroup === YOUNG_ADULT_CONSENT;
+};
+
+const isUndefined = (arg: any): arg is undefined => arg === undefined;
+
+// TODO: decide if participant contact fields will be excluded in a guardian is present
+// TBD in https://github.com/OHCRN/platform/issues/388
+export const hasRequiredParticipantContactInfo = (
```
joneubank, line 47: This function has mystery logic. A TSDoc comment would be helpful.

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+export const hasRequiredGuardianInformation = (
+	props: {
+		consentGroup: ConsentGroup;
+	} & GuardianBaseFields,
+) => {
+	// guardianName, guardianPhoneNumber, guardianEmailAddress, guardianRelationship must be defined if
+	// GUARDIAN_CONSENT_OF_MINOR, GUARDIAN_CONSENT_OF_MINOR_INCLUDING_ASSENT or ADULT_CONSENT_SUBSTITUTE_DECISION_MAKER was selected
+
```
joneubank, line 65: Looks like a TSDoc description to me!

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+		consentGroup,
+		guardianName,
+		guardianPhoneNumber,
+		guardianEmailAddress,
+		guardianRelationship,
+	} = props;
+
+	return requiresGuardianInformation(consentGroup)
+		? [guardianName, guardianPhoneNumber, guardianEmailAddress, guardianRelationship].every(
+				(field) => !isUndefined(field),
```
joneubank, line 76:
```ts
const hasValue = <T>(input: T | undefined): input is T => !isUndefined(input);

// ...

someArray.every(hasValue);
```

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -26,86 +26,82 @@ import { PIClinicianInviteResponse } from '../../src/services/piDas/index.js';
...
-				participantLastName: 'Simpson',
-				participantEmailAddress: 'bart.simpson@example.com',
-				participantPhoneNumber: '6471234567',
-				consentGroup: ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR,
-				guardianName: 'Marge Simpson',
-				guardianRelationship: 'Wife', // missing guardianEmailAddress and guardianPhoneNumber
-				consentToBeContacted: true,
-			}).success,
-		).false;
+		const result = ClinicianInviteBase.safeParse({
```
joneubank, line 71: This is a good use of test over zod schema, making sure a specific refinement is being enforced.

Perhaps a test fixture on the common refinement functions themselves would be useful. Maybe that already exists...

packages/types/tests/entities/ParticipantIdentity.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+			ohipNumber: '1234567890',
+			currentLifecycleState: LifecycleState.enum.REGISTERED,
+			participantPreferredName: 'Homer',
+			participantOhipFirstName: 'Homer',
+			participantOhipLastName: 'Simson',
+			dateOfBirth: new Date(),
+			participantPhoneNumber: '6471234567',
+			residentialPostalCode: 'T4B0V7',
+			participantEmailAddress: 'homer.simpson@example.com',
+			consentGroup: ConsentGroup.enum.GUARDIAN_CONSENT_OF_MINOR, // missing all guardian contact fields
```
joneubank, line 87: These comments are incredibly helpful. Thank you.

joneubank (commented):
Loving the code patterns here in general. Was pretty clear to read through the work. I have left you a bunch of thoughts for places to do a bit of improvement.

apps/data-mapper/src/services/das/consent.ts
```diff
@@ -212,3 +214,33 @@ export const getParticipantResponsesByQuestionId = async ({
...
+	try {
+		const result = await axiosClient.post(urlJoin(consentDasUrl, 'participants'), req);
+		return success(result.data.participant);
+	} catch (error) {
+		if (error instanceof AxiosError && error.response) {
+			const { data, status } = error.response;
+
+			if (status === 400) {
+				logger.error('Invalid request while creating participant', data);
+				return failure('INVALID_REQUEST', data.message);
```
mistryrn, line 236: Looks like this may be the case where `INVALID_REQUEST` is being returned by the `createParticipant` function, so we may need to add a case to the switch statement after all

apps/data-mapper/src/services/das/keys.ts
mistryrn: I can clarify these @joneubank  😅 They're from the original Data Mapper POC many months ago and Ann is just moving my hasty POC code around here. We haven't properly implemented these yet, they will get addressed in the coming sprints when we work on this functionality

TL;DR: it's all my fault, but we'll be fixing it soon 🙈

packages/types/src/common/conditionalFieldUtils.ts
anncatton: aha that is prob what i was missing! i had tried `includes` and got TS yelling, but admittedly got lazy and just moved on

packages/types/tests/entities/ClinicianInvite.spec.ts
anncatton: i think at the moment our tests our only with the refinement applied to specific schemas using it. Will try this out.

packages/types/tests/entities/ParticipantIdentity.spec.ts
anncatton: i think the credit here goes to Alyssa 🙂

apps/data-mapper/src/services/das/keys.ts
anncatton: in my mind i had put a note about this in the PR description - but i DIDN'T!! my bad. but yes, these will be implemented properly when the services are. I just wanted to get these functions into the proper files for now.

apps/data-mapper/src/services/das/pi.ts
anncatton: I'm also not sure it's necessary - the reason i thought of this roll back at all is that PI DAS schema is not up to date with the schema types changes i have in this pr, so testing locally the pi das response was automatically invalid. (I've a ticket created to implement the field changes in the DAS: https://github.com/OHCRN/platform/issues/391 )

Just to make sure I understand your suggestion - we rely on the PI DAS itself to validate its response data (and rollback if necessary), so that Data Mapper wouldn't be responsible for that portion of the transaction?

We do have this pattern elsewhere in this PR (to delete pi data if consent participant step fails) and in the Clinician Invite flow. Sounds like maybe we should revisit?

mistryrn (commented):
Lots of great stuff here, really liking the new types structure/organization!

apps/data-mapper/src/services/das/consent.ts
anncatton: i think what i have here is incorrect - i forgot to handle the conflict error instead, so this should be checking for a `409` from the consent das (like i have in the pi create service call). The `withRequestValidation` wrapper on the router `POST` will return a `400` if the request body doesn't validate against the `CreateParticipantRequest` schema, so we shouldn't need the explicit case as I have it here.

packages/types/src/entities/ParticipantIdentity.ts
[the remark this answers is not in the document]
anncatton: yeah maybe i can bring this up in our demo meeting!

apps/data-mapper/src/routers/participants.ts
```diff
@@ -46,19 +49,59 @@ router.get('/:participantId', async (req, res) => {
...
+ *             $ref: '#/components/schemas/CreateParticipantRequest'
+ *     responses:
+ *       201:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/CreateParticipantResponse'
+ *       400:
+ *         description: RequestValidationError - The request body was invalid.
```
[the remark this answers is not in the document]
anncatton: we don't, but sounds like a nice idea!
I will add an enhancement ticket for this to avoid making this PR larger than it needs to be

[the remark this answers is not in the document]
anncatton: https://github.com/OHCRN/platform/issues/399

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -29,40 +29,65 @@ const {
...
 // TBD in https://github.com/OHCRN/platform/issues/388
+/**
+ * Checks if a Participant schema object contains the required Participant contact fields needed for the ConsentGroup
+ *
+ * participantEmailAddress and participantPhoneNumber must be defined if ADULT_CONSENT or YOUNG_ADULT_CONSENT was selected
+ * @param {Object} props
+ * @param {ConsentGroup} props.consentGroup - the ConsentGroup associated with a Participant object
+ * @param {string} props.participantEmailAddress - value for participantEmailAddress in a Participant object
+ * @param {string} props.participantPhoneNumber - value for participantPhoneNumber in a Participant object
+ * @returns {boolean} returns true if all required fields are present
```
joneubank, line 58: To help simplify... I don't think we need to declare the types of variables in the TSDocs. They are prone to becoming out of sync with the actual code, and if that happens I'll trust the TS interpretter over the docs anyways. The descriptive text is really the win here.

joneubank: ... In this case, we might not even need the description of the params. We know what a consentGroup.participantPhoneNumber is I think.

joneubank: The return description (true if all fields are present) is real helpful though

packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -29,40 +29,65 @@ const {
...
 };
 
 const requiresParticipantContactInfo = (consentGroup: ConsentGroup) => {
-	return consentGroup === ADULT_CONSENT || consentGroup === YOUNG_ADULT_CONSENT;
+	const consentOptions: ConsentGroup[] = [ADULT_CONSENT, YOUNG_ADULT_CONSENT];
+	return consentOptions.includes(consentGroup);
 };
 
 const isUndefined = (arg: any): arg is undefined => arg === undefined;
+const hasValue = <T>(input: T | undefined): input is T => !isUndefined(input);
```
joneubank: There's no right answer to this question but: are these the kind of functions we are likely to want to reuse? Should they be moved somewhere to make them available in other files? other modules? How many instances of these functions are future devs going to make? Maybe thats ok?

packages/types/tests/utils/conditionalFields.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+		YOUNG_ADULT_CONSENT, // non-guardian group
+	} = ConsentGroup.enum;
+	describe('hasRequiredGuardianInformation', () => {
+		it('returns TRUE for a guardian consentGroup (GUARDIAN_CONSENT_OF_MINOR) and all required guardian fields are provided', () => {
+			const guardianFields = {
+				guardianName: 'Gina Guardian',
+				guardianEmailAddress: 'gina_g@example.com',
+				guardianPhoneNumber: '0123456789',
+				guardianRelationship: 'Mother',
+			};
```
joneubank, line 43: This object is repeated a lot, you could define once at top of the file and reuse it.

packages/types/tests/utils/conditionalFields.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+			expect(result).false;
+		});
+
+		it('returns TRUE for a non-guardian consentGroup and no guardian fields are provided', () => {
+			const testSchemaObj = {
+				consentGroup: ADULT_CONSENT,
+			};
+			const result = hasRequiredGuardianInformation(testSchemaObj);
+			expect(result).true;
+		});
```
joneubank, line 122: Is there an inverse case for this? What do we return for a non-guardian when consentGroups ARE provided.

joneubank (dismissed):
Looks great Ann, thanks.

I know it's not always possible but breaking this into smaller PRs is a good objective for us all.

packages/types/src/common/conditionalFieldUtils.ts
anncatton: yeah the other thing about that is that Intellisense doesn't actually show the type in the hover definition, if following the `{type}` syntax. Not sure if this is a problem with VSCode or jsdoc!

anncatton: i made a little compromise and left in `@params props prop1, prop2, ...`

mistryrn (approved):
🚀

Issue #348, opened 2024-01-05
Add Special UI Component for OHIP Field


b-f-chan:
Add a special UI component to handle the OHIP field on Step 2 of the Consent Wizard

This is a custom component where there is a text field with an associated checkbox input that has conditional logic

Please see mockups for reference

@samrichca to add any details required

---

- Make a stub form on an unused page that will eventually use this component (form context is needed to create an input)
- I don't think any new input components are required
- Add a new fieldset for a text input that can be disabled with a checkbox
- Add disabled & error states

chelseagmarie:
Hi, where is it possible to find the mockups to reference for this issue?

xusandrew:
blocked - need flag for if the OHIP field disabled/enabled in data model

b-f-chan:
Merged, closing

PR #395, opened 2024-01-24
💄#348: OHIP component

xusandrew:
## Description of Changes

Created special React component for the OHIP input field, found on consent wizard step two. Comes with disabled + error states.

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+	const OHIPFieldId = useId();
+	const OHIPTextInputId = `${OHIPFieldId}-text`;
+	const OHIPCheckboxId = `${OHIPFieldId}-checkbox`;
+
+	const { watch, setValue } = useFormContext();
+	const checkboxValue: boolean = watch(OHIPCheckboxId);
+
+	useEffect(() => {
+		if (checkboxValue) {
+			setValue(name, '' as any);
```
xusandrew, line 62: 😿dodging a type error here that I was blocked on:

Argument of type 'string' is not assignable to parameter of type 'Path<T> extends `${infer K}.${infer R}` ? K extends string ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? R extends `${infer K}.${infer R}` ? K extends string | number | symbol ? R extends string ? any : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends`${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : R extends string | number | symbol ? any : R extends `${number}` ? unknown : never : never : K extends `${number}` ? never : never : Path<T> extends string ? any : Path<T> extends `${number}` ? never : never'.ts(2345)

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -305,6 +306,18 @@ const ClinicianInviteFormComponent = ({
...
+						label={'10-digit number on front of card'}
+						name="clinicianInstitutionalEmailAddress"
+						required
+						// tooltipContent={textDict.clinicianInstitutionalEmailAddressTooltip}
+						// type="email"
+						withNarrowDesktopLayout
+						title={'OHIP #'}
+						subtitle={'(Green Health Card)'}
+						checkboxLabel={'I do not have an Ontario health card.'}
+					/>
```
xusandrew, line 320: just for testing, will remove before merge

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
xusandrew: y is it so long

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}, [checkboxValue, name, setValue]);
+
+	return (
+		<fieldset
```
samrichca, line 67: is it possible to use the FieldSet component that comes with a tooltip, similar to how the SelectFieldSet is implemented? if it doesn't look quite right you can modify the FieldSet or pass down a classname/style

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+					<h4 className={styles.subtitle}>
+						{subtitle}
+						{required && !title && <RequiredAsterisk />}
+					</h4>
+				)}
+			</div>
+
+			<div className={styles.OHIPWrapper}>
+				<div className={styles.OHIPTextWrapper}>
+					<TextInput
```
samrichca, line 92: this input needs a label

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -305,6 +306,18 @@ const ClinicianInviteFormComponent = ({
 						required
 						withNarrowDesktopLayout
 					/>
+					<OHIPFieldSet
+						error={errors.clinicianInstitutionalEmailAddress?.type && errorsDict.required}
+						label={'10-digit number on front of card'}
+						name="clinicianInstitutionalEmailAddress"
+						required
+						// tooltipContent={textDict.clinicianInstitutionalEmailAddressTooltip}
+						// type="email"
```
samrichca, line 315: OHIP number shouldn't use the type="email"

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
xusandrew: I will try this first thing tmr! 

I initially decided against using the FieldSet component though -- since the FieldSet component only supports a singular label, while the OHIP field requires like 3:
 
(from the mockups)
- OHIP # 
- (Green Health Card) 
- 10-digit number on front of card

Modifying the FieldSet component to support this seemed like it would be unnecessary complexity. Plus, this OHIP field is the only component that will be structured this way, might be better just to leave it.

Let me know what you think though :)

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+						className={styles.OHIPTextInput}
+						disabled={disabled || checkboxValue}
+						id={OHIPTextInputId}
+						name={name}
+						required={required}
+						type={'text'}
+					/>
+
+					{label && (
+						<label className={styles.label} htmlFor={OHIPTextInputId}>
```
samrichca, line 102: inputs have to always have a label, so labels shouldn't be conditionally rendered. I would make the label element the "OHIPHeader" div instead for consistency with the other fields that look similar, since labels are interactive (i.e. when a user clicks on them, the input will receive focus)

aria-describedby attributes accept a list of IDs separated by spaces - you're right to associate this element with the input, but I would do it as a description instead

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
samrichca: ok no problem. I would like to reuse CSS and tooltips somehow from the main fieldset, but we're going to overhaul tooltips next sprint, so I'm OK with this for now to get moving on step 2 👍

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+	const OHIPCheckboxId = `${OHIPFieldId}-checkbox`;
+
+	const { watch, setValue } = useFormContext();
+	const checkboxValue: boolean = watch(OHIPCheckboxId);
+
+	useEffect(() => {
+		if (checkboxValue) {
+			setValue(name, '' as any);
+		}
+	}, [checkboxValue, name, setValue]);
```
mistryrn, line 64: What is the intention of the `useEffect` here?

xusandrew: When the checkbox is checked, it deletes the current value in the textfield

mistryrn: Ah, I see. Gotcha. Out of curiosity, have you tried it without the `useEffect`? I'm seeing the value getting cleared even without it – not sure if toggling the `disabled` state is automatically clearing the value for us? Could be worth digging into the react-hook-forms docs to see if that's the case. Might be able to avoid the `any` and thus the big types error altogether if that's the case 🙂

xusandrew: ill give it a try!

xusandrew: Very interesting, it seems like I dont need this useEffect at all. I did some digging on the docs to try and find why,  cant seem to find anything though. Would like to understand how this fully works before implementing this though, I'll do some more digging tommorow.

samrichca:
I tried using this locally and had a couple of issues

font size inside the input is too small - it has to be 1rem/16px or mobile Safari will zoom in when the user taps on the field
if I enter a number in the OHIP field, then click on another input, my OHIP number disappears

xusandrew:
@samrichca 

Fixed the font sizing, and the reason why this field was getting deleted was because it shared the same "name" attribute as the field below, which is "clinicianInstitutionalEmailAddress". My bad there, it should now work fine :)

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
xusandrew: I finally FOUND IT, turns out the useEffect is required.

When I used this component I passed in the name="clinicianInstitutionalEmailAddress", when there was already a TextField with name="clinicianInstitutionalEmailAddress" right under it. That is what messed with the value inside of the field when the state updated. Now, I've commented out the other textfield, and when I click the checkbox now the value remains:

Screenshot 2024-01-30 at 2 34 25 PM

But it works with the useEffect:

Screenshot 2024-01-30 at 2 34 51 PM

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -315,15 +316,27 @@ const ClinicianInviteFormComponent = ({
...
+						withNarrowDesktopLayout
+						title={'OHIP #'}
+						subtitle={'(Green Health Card)'}
+						checkboxLabel={'I do not have an Ontario health card.'}
+					/>
+					{/* <TextFieldSet
```
xusandrew, line 331: remove comment before merge

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
samrichca: image

it throws an error if I check the box and submit the form 🤔

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
mistryrn, line 64: Spent some more time this morning digging into this type error so we can avoid using `any`. I didn't realize react-hook-form offers a `resetField` function as well: https://react-hook-form.com/docs/useform/resetfield

I've only done some basic testing with it, so you'll want to double check/do your own local testing, but it seems to be resetting the value nicely with the added benefit of not requiring us to specify the value to set the field to (and thus avoiding having to specify the type of that value 😛 )

```suggestion
	const { watch, resetField } = useFormContext();
	const checkboxValue: boolean = watch(OHIPCheckboxId);

	useEffect(() => {
		if (checkboxValue) {
			resetField(name);
		}
	}, [checkboxValue, name, resetField]);
```

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
mistryrn: Good catch, @samrichca . @xusandrew we'll probably want to look at the `required` props on the `TextInput` and `CheckboxInput` and rather than just passing `required` to them, do something like:
```ts
// TextInput
// required if checkboxValue is false (i.e. checkbox is not checked)
required={checkboxValue ? false : required}

// CheckboxInput
// required if checkboxValue is true (i.e. checkbox is checked
required={checkboxValue}
``` 

Note that those probably aren't the right conditions, but something along those lines will hopefully do the trick 🙏

mistryrn (commented):
Component looks great! Thanks for looking into the clear/reset behaviour, I may have found a solution to that big type error.

Other than that, the only thing I noticed was the `required` behaviour that Sam mentioned. Suggested a potential solution but it'll need some work

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/OHIPFieldSet.module.scss
```diff
@@ -0,0 +1,108 @@
+.OHIPFieldset {
```
mistryrn, line 1: One minor thing – for consistency let's rename these class names to `lowerCamelCase` 🙂

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
samrichca: you're on the right track. the required attribute on inputs will signal to screenreaders that these inputs are required.

for validation purposes you'd have to add a `refine` chain to the zod schema - lots of examples in the codebase. this only works in onSubmit validation though. not sure about onBlur yet but @xusandrew you're welcome to investigate

apps/consent-ui/src/components/common/Form/fieldsets/OHIPFieldSet/index.tsx
xusandrew: Good find! This is exactly the usecase I would want a function like that for. Made the change

xusandrew:
because of the change in the mocks, I was able to make this component a lot cleaner!

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -302,14 +303,27 @@ const ClinicianInviteFormComponent = ({
...
 						error={errors.clinicianInstitutionalEmailAddress?.type && errorsDict.required}
 						label={labelsDict.clinicianInstitutionalEmailAddress || ''}
 						name="clinicianInstitutionalEmailAddress"
 						required
 						description={textDict.clinicianInstitutionalEmailAddressTooltip}
 						type="email"
-					/>
+						withNarrowDesktopLayout
```
xusandrew, line 325: will revert the change here when merging - it is here just for testing now

mistryrn (dismissed):
Changes look good! Let's hold off on merging until Sam is back for final approval though 🙂

apps/consent-ui/src/components/views/Invite/ClinicianInviteForm/index.tsx
```diff
@@ -305,6 +306,18 @@ const ClinicianInviteFormComponent = ({
 						required
 						withNarrowDesktopLayout
 					/>
+					<OHIPFieldSet
+						error={errors.clinicianInstitutionalEmailAddress?.type && errorsDict.required}
+						label={'10-digit number on front of card'}
+						name="clinicianInstitutionalEmailAddress"
+						required
+						// tooltipContent={textDict.clinicianInstitutionalEmailAddressTooltip}
```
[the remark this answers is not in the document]
samrichca: these 2 lines are still here 👀

samrichca:
please fix the merge conflict & i'll approve ASAP 👍 🚀

xusandrew:
@samrichca should be okay to merge now!

samrichca (approved):
lgtm! 🚀

Issue #362, opened 2024-01-15
Consent Wizard: Progress header items don't have a unique key


samrichca:
In apps/consent-ui/src/components/common/ProgressHeader/index.tsx on line 49, the `map` returns a fragment, which doesn't have a unique key. This generates a "Warning: Each child in a list should have a unique "key" prop." error in the terminal. Converting the fragment to a div with a unique key breaks the CSS.

- Change the fragment to a div with a unique key
- Fix the resulting CSS issues

Exit Criteria:
- Page looks exactly like it does now
- Unique key error is resolved

b-f-chan:
@xusandrew can this be closed?

xusandrew:
@b-f-chan yes

b-f-chan:
Reviewed & merged --> Closing

PR #418, opened 2024-02-05
🐛#362: Fix unique key warning in consent wizard header

xusandrew:
## Description of Changes

Fix unique key warning in consent wizard header.

Changing the Fragment to a Div in this scenario would have introduced a lot more complexity with the styling, so I decided against that. Instead, I found this solution which is clean, and also removes the warning.

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/components/common/ProgressHeader/index.tsx
```diff
@@ -46,7 +47,7 @@ const ProgressHeader = async ({
 			{/* Tablet and Desktop screens */}
 			<div className={styles.header}>
 				{steps.map((step, index) => (
-					<>
+					<Fragment key={index}>
```
samrichca, line 50: good call on using `fragment`, but indices shouldn't be used for keys. see the "pitfall" callout here: https://react.dev/learn/rendering-lists#why-does-react-need-keys

apps/consent-ui/src/components/common/ProgressHeader/index.tsx
```diff
@@ -46,7 +47,7 @@ const ProgressHeader = async ({
...
 						<div className={styles.step} key={step.name}>
```
samrichca, line 51: `key={step.name}` can be removed from the div

samrichca (approved):

Issue #366, opened 2024-01-16
Register Form: Implement radio & date inputs


samrichca:
Currently in the registration form there are placeholders for the radio & date inputs, because these inputs haven't been built yet.

- Add radio input to "are you registering on behalf of someone else" field
- Add date input to "date of birth" field
- Update type & Zod schema for the form
  - This isn't implemented in types/consentApi yet, so continue to use the stubs in the UI.
- Make sure the form works
- Make substitute decision maker fields optional depending on radio input values
- Show "this participant is a minor" modal depending on date of birth selected

Exit Criteria:
- Form has radio inputs for the first question
- Form has a date input for date of birth
- If the user decides to use a substitute decision maker, show the fields for substitute decision maker info
- If the user enters a date of birth for a participant younger than 18, show a modal directing them to talk to a clinician

## Blocked

- Radio & date inputs not done
- Register form not done

b-f-chan:
Component PRs reviewed, fixed, merged this ticket is no longer blocked, removing label

b-f-chan:
Will open PR for date portion of this and wait to add Guardian fields later

samrichca:
Separate ticket for modal https://github.com/OHCRN/platform/issues/456

PR #441, opened 2024-02-14
💄 #366 1 of 2: add date of birth to register page

samrichca:
### Consent UI
- Update calendar input
  - Add onBlur validation
  - Add aria-required
- Add date of birth field to Register page & update schema
- No modal for date of birth validation. Will be covered in https://github.com/OHCRN/platform/issues/456

## Types
- Add date-fns
- Add age check function
  - Add tests

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/components/views/Register/RegisterForm/FormStep1.tsx
```diff
@@ -75,6 +81,14 @@ const FormStep1 = ({
 		setFocus('guardianName');
 	}, [setFocus]);
 
+	const handleDateOfBirthBlur = () => {
+		const dateOfBirthValue = getValues('dateOfBirth');
+		const userIsAdult = checkAge18AndOver(dateOfBirthValue);
+		if (!userIsAdult) {
+			console.log('participant is a minor');
```
samrichca, line 88: I have to add the modal from the modal refactor PR.

apps/consent-ui/src/components/views/Register/RegisterForm/types.ts
```diff
@@ -37,13 +38,22 @@ export const RegisterFormStep1 = z.object({
 	// commenting this out because the form won't work
 	// with unused fields in the Zod schema
 });
+
+const DateOfBirthField = z
```
samrichca, line 42: split up date of birth into its own schema because it has a refinement

apps/consent-ui/src/components/common/Form/fieldsets/inputs/CalendarInput.tsx
```diff
@@ -54,22 +54,27 @@ const CalendarInput = <T extends FieldValues>({
...
-					locale={currentLang}
-					onBlur={onBlur}
-					onFocus={onFocus}
-					disabled={disabled}
-					{...ariaProps}
-				/>
-			)}
+			render={({ field: { onBlur: controlOnBlur, onChange, value } }) => {
+				const handleBlur = () => {
+					controlOnBlur();
```
samrichca, line 59: - pass controlOnBlur to the input, so blur validation works
- use onBlur handler from the prop as well, to show the "user is a minor" modal

apps/consent-ui/package.json
```diff
@@ -30,6 +29,7 @@
 		"tailwindcss": "3.3.2",
 		"types": "workspace:^",
 		"url-join": "^5.0.0",
+		"use-onclickoutside": "^0.4.1",
```
samrichca, line 32: this library uses onMouseDown events instead of onClick. the event order is mousedown, blur, click - if we open a modal on blur, we shouldn't close it with the click event that happens immediately after.

apps/consent-ui/src/components/views/Register/RegisterForm/types.ts
```diff
@@ -38,13 +38,22 @@ export const RegisterFormStep1 = z.object({
...
+	.object({
+		dateOfBirth: z.coerce.date(),
+	})
+	.refine(checkParticipantIsAdult, { message: 'participantIsMinor', path: ['dateOfBirth'] });
```
samrichca, line 46: second argument is needed to show an error on this field

apps/consent-ui/src/components/views/Register/RegisterForm/types.ts
```diff
@@ -38,13 +38,22 @@ export const RegisterFormStep1 = z.object({
...
 export type RegisterFormStep1 = z.infer<typeof RegisterFormStep1>;
 
-export const RegisterFormStep2Fields = z.object({
+// STEP 2
+
+const RegisterFormStep2Fields = z.object({
 	consentToBeContacted: z.literal(true),
 	participantEmailAddress: z.string().email(),
 });
-export type RegisterFormStep2Fields = z.infer<typeof RegisterFormStep2Fields>;
```
samrichca, line 47: these don't need to be exported, and the type isn't needed

apps/consent-ui/src/components/common/Form/fieldsets/inputs/CalendarInput.tsx
```diff
@@ -54,20 +52,19 @@ const CalendarInput = <T extends FieldValues>({
 		<Controller
 			control={control}
 			name={name}
-			render={({ field: { onChange, value } }) => (
+			render={({ field: { onBlur, onChange, value } }) => (
```
samrichca, line 55: onBlur handler is needed for validation

apps/consent-ui/src/components/common/Form/fieldsets/inputs/CalendarInput.tsx
```diff
@@ -40,8 +40,6 @@ const CalendarInput = <T extends FieldValues>({
 	popperClassName,
 	id,
 	name,
-	onBlur = () => {},
```
samrichca, line 43: these aren't used

apps/consent-ui/src/components/views/Register/RegisterForm/types.ts
```diff
@@ -17,15 +17,15 @@
...
 import { NonEmptyString } from 'types/common';
-import { Name, PhoneNumber, hasMatchingPasswords } from 'types/entities';
 import { z } from 'zod';
 
 // TODO hookup backend #368
 // create a better zod schema with conditional validation,
 // and optional name fields
 
-export const RegisterFormStep1 = z.object({
+const RegisterFormStep1Fields = z.object({
```
anncatton, line 28: I tried running this locally and am not able to get the date validation to succeed. Is that expected?
I am wondering if there is a conflict here with the `dateOfBirth` field, as it's typed as a `NonEmptyString` here and then added as the separate date field in the `DateOfBirthField` type. Intellisense is showing them as 2 separate fields in the intersection. I commented out the field in `RegisterFormStep1Fields` and got a valid birthdate to pass validation on the form.

I know `merge()` will overwrite field types with the same key, but it appears `intersection()` does not behave the same way (and you have to use the latter because of the refinement)

packages/types/src/entities/fields/DateOfBirth.ts
```diff
@@ -0,0 +1,38 @@
+/*
```
anncatton, line 1: Since these are just utility functions i would move them out of the `fields` folder. Maybe we should be adding a `utils` dir to the `common` folder for these kinds of functions. We have the `conditionalFieldIUtils` file in `common` as well, but since we are creating more refinement functions, we might need to reorganize that a little so we don't end up with a giant utils file.
Maybe something like this?
```
.
└── common/
    └── utils/
        ├── guardian.ts
        ├── participant.ts
        └── dateOfBirth.ts
```
If you like that structure, you can just move these to the `dateOfBirth` file in this new dir. I can reorganize the `conditionalUtils` in https://github.com/OHCRN/platform/issues/450

packages/types/tests/entities/DateOfBirth.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { describe, expect, it } from 'vitest';
+
+import { checkAge18AndOver } from '../../src/entities/fields/index.js';
+
+describe('Date of Birth', () => {
```
anncatton, line 24: Thanks for adding these tests! I would love to see an example test for how your separate refined field works with the whole schema validation, i.e. running a validation on the whole registerFields schema with an invalid dob would return the custom refinement error, not just `success === false`

apps/consent-ui/src/components/views/Register/RegisterForm/types.ts
samrichca: good catch!

packages/types/tests/entities/utils/DateOfBirth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { describe, expect, it } from 'vitest';
+
+import {
+	checkIsMinimumAgeOrGreater,
+	MINIMUM_AGE_IN_YEARS,
+} from '../../../src/common/utils/index.js';
+
+describe('Date of Birth', () => {
+	const startDate = new Date('02/28/2024');
```
anncatton, line 28: 👍

anncatton (commented):
This looks great Sam! Thanks for creating the ui service folder in types. Just one last change to ask for, can you move your `dateOfBirth.spec` file into the `/tests/utils/` folder, instead of inside `/tests/entities/utils/`?

anncatton (dismissed):

packages/types/tests/utils/dateOfBirth.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { describe, expect, it } from 'vitest';
+
+import { checkIsMinimumAgeOrGreater, MINIMUM_AGE_IN_YEARS } from '../../src/common/utils/index.js';
+
+describe('Date of Birth', () => {
+	const startDate = new Date('02/28/2024');
```
mistryrn, line 25: Is there a possibility that the tests below might get out of sync over time since we're using a fixed date as the `startDate` here? I'm not sure but by next year would the `exactlyMinimumAgeDateOfBirth` one fail because it'll be doing `2024 - MINIMUM_AGE_IN_YEARS` instead of `2025 - MINIMUM_AGE_IN_YEARS`?

anncatton: Good catch @mistryrn , i missed that little tidbit! - yes, for these tests i think we should have all fixed dates, so maybe we can mock the date the util function is using.

samrichca: I might be able to pass down the current date when I call the refinement function, instead of getting the current date in the refinement function, which would allow us to have mock fixed dates for tests

mistryrn (approved):
Updated tests look great! Thanks for getting all these changes in 🙂

Issue #167, opened 2023-10-24
Build consent API endpoint --> Header & Dashboard: `GET /user`


b-f-chan:
Build Consent API Endpoint: `GET /user`

**Applicable Page:** Header on all pages, participant dashboard page
**Functionality:** Check participant's
- firstName
- lastName
- email
- role
- emailVerified
- enabled2FA

**Expected Outcome(s):**

- [ ] Swagger doc for the endpoint is functional and testable
- [ ] Unit testing implemented for this endpoint 

---

Update 10/30/2023 from Sam:
- Use MOCK_USER_ID
- Endpoint mapped out in this PR https://github.com/OHCRN/platform/pull/176

b-f-chan:
Ignoring `role` field for Consent Portal MVP, add later to DB if needed

b-f-chan:
Done, closing

PR #508, opened 2024-03-22
🛂 371 - create auth middleware to parse bearer token from headers

anncatton:
## Description of Changes

Adds an authentication middleware that parses a Bearer token from incoming requests, and if the token is valid, adds the token `sub` to the request object for consumption by the endpoint. The `sub` is the `keycloakId` and can be used to query for the participant.

### Consent API

- adds `jsonwebtoken` for decoding and verifying JWTs
- adds a `withAuth` middleware to parse any Bearer token provided in the `Authorization` header. Tokens must be prepended with `Bearer` to pass validation. The `jwt.verify` function will error on a missing, malformed or expired token. 
- there is currently no permissions-based access in the consent api, so any failed validation will return a `401 Unauthorized`
- splits the app config into 2 separate objects, for "static" values from `process.env`, and an AuthConfig that retrieves the Keycloak public key from the applicable Keycloak realm. Once initialized, the Authconfig should not need to retrieve the key again (fetch is intended to occur on startup only). 
- The public Key can also be set in `process.env`.
- server start will fail if no value is provided for the public key.
- for the purposes of this ticket, the keycloak public key is assumed present from the config setup; I tested with an env var for convenience. There will be a followup task to retrieve the public key from the Keycloak realm and set the value on the service start up.
- the middleware can be added as needed to the router endpoints that require authorized access. For testing purposes, I've added it to the `GET /user` endpoint, with a mock success response. Need to decide what data needs to be returned from this endpoint, and can it all be retrieved from the Keycloak access token?
- the global type declaration is needed to allow adding custom data to the request object passed from the middleware - so we can add in data as needed (as recommended in @joneubank 's [example PR](https://github.com/OHCRN/platform/pull/506) )

### Types

- Adds `UnauthorizedErrorResponse` type and enum

### Special Instructions

Before running these changes, you will need to install `jsonwebtoken` and supporting types
```
pnpm i
```

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-api/src/config.ts
```diff
@@ -11,6 +11,10 @@ export const getAppConfig = () => {
 		recaptcha: {
 			secretKey: process.env.RECAPTCHA_SECRET_KEY || 'MISSING_RECAPTCHA_SECRET_KEY',
 		},
+		// TODO: retrieve public key from keycloak and save on server start
+		auth: {
```
anncatton, line 15: For now the middleware will return a 500 server error if the key is missing, can the server fail if the key isn't retrieved + set?

apps/consent-api/src/routers/participants.ts
```diff
@@ -86,18 +86,19 @@ router.get('/', async (req, res) => {
...
  */
-router.get('/:id', async (req, res) => {
-	const { id } = req.params;
-	try {
-		const participant = await getParticipant(id);
-		return res.status(200).send({ participant });
-	} catch (error) {
-		logger.error(error);
-		return res.status(404).send({ error: 'Participant not found' });
+router.get('/:id', withAuth, async (req, res) => {
```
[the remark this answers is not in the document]
mistryrn: Yeah I get what you're saying here. Almost feels more like a `GET /users` but not quite. Technically it is still the Participant as the entity, but we're getting it by `keycloakId` so 🤔 

Will have to think about this for sure!

mistryrn (commented):
Looks great! Will have to think about that naming and also the handling of the keycloak public key in DEV/QA/etc, but I like this approach! Very clean stuff 🙂

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const logger = serviceLogger.forModule('Auth');
+// eslint-disable-next-line import/no-named-as-default-member
+const { verify } = jwt;
+
+declare global {
+	// eslint-disable-next-line @typescript-eslint/no-namespace
+	namespace Express {
+		interface Request {
+			keycloakId?: string;
```
joneubank, line 35: this is the ID of the keycloak? or is it the subject ID taken from the keycloak auth token?

Feels like ambiguous naming.

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,98 @@
...
+	if (typeof decoded == 'string' || decoded === null) {
+		throw Error('Invalid JWT format');
+	} else {
+		return decoded;
+	}
+};
+
+const parseBearerToken = (token?: string): string => {
+	return token && token.startsWith('Bearer ') ? token.replace('Bearer ', '') : '';
+};
```
joneubank, line 54: These seem like common utils that could be made generally available. No reason to overly refactor before there is a need, but you should keep an eye open as you keep coding and reviewing code for other places where these operations are done. Its possible we'd want to add these to a common package in the future so we can reuse this logic.

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Router middleware to ensure the incoming request includes a valid access token in the Authorization header.
+ * When added to a router, it is expected that the incoming request has a header containing a valid access token
+ * in the JWT Bearer format. Any request missing this token or with an invalid token will be rejected with an HTTP 401 status.
+ *
+ * If the token is valid, the "sub" value from the decoded token will be added to the Request object as the "keycloakId"
+ *
+ * @example
```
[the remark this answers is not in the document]
joneubank: `@example` is a great feature of JSdoc comments.

Thanks for JS doc on the middleware @anncatton

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,98 @@
...
+	try {
+		const verifiedToken = decodeToken(keycloakPublicKey, accessToken);
+		// add the sub to the Request obj as keycloakId
+		req.keycloakId = verifiedToken.sub;
+		return next();
+	} catch (err: unknown) {
+		if (err instanceof Error) {
+			if (err.cause === 'missingPublicKey') {
+				logger.error('Unexpected error', err);
+				return res.status(500).json(ServerErrorResponse('Unexpected error'));
```
joneubank, line 89: Is this the correct error message? This is an auth error, the error is expected, its caused by the bearer token being missing.

apps/consent-api/src/routers/participants.ts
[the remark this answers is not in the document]
joneubank: If you are getting the `keycloakId` from the Authorization token, is it not just a request to GET self? Like, you arent actually requesting by ID, you are just able to retrieve information related to the user indicated in the provided token.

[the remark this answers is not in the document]
anncatton: ya exactly...i used this existing endpoint for the sake of convenience, but it was created when we were still just picturing the most straightforward REST setup. 
We do actually have a `/user` router which i forgot existed...maybe i can move this initial JWT setup there https://github.com/OHCRN/platform/blob/develop/apps/consent-api/src/routers/user.ts#L34

apps/consent-api/src/middleware/auth.ts
anncatton: yeah good point...at this stage it is the `sub` from the jwt. it becomes `keycloakId` in the ohcrn backend context

apps/consent-api/src/config.ts
anncatton: Server will fail to start if the `keycloakPublicKey` is an empty value

apps/consent-api/src/authConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+dotenv.config();
+let authConfig: AuthConfig | undefined;
+
+const fetchPublicKey = async (path: string): Promise<string> => {
+	try {
+		const response = await fetch(path);
+		const result = await response.json();
+		return `-----BEGIN PUBLIC KEY-----\n${result.public_key}\n-----END PUBLIC KEY-----`;
+	} catch (err) {
+		console.error(err);
```
mistryrn, line 37: Should we be using the `logger` here instead of `console`?

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,92 @@
...
+	const authHeader = req.headers.authorization;
+	const accessToken = parseBearerToken(authHeader);
+	const {
+		auth: { keycloakPublicKey },
+	} = await getAppConfig();
+
+	try {
+		const verifiedToken = decodeToken(keycloakPublicKey, accessToken);
+		// add the sub to the Request obj as keycloakId
+		req.keycloakSubjectId = verifiedToken.sub;
```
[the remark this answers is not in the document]
mistryrn: Thanks for calling this out. Should we see how we get on with the registration flow and then make a decision on adding this?

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -29,7 +29,7 @@ const logger = serviceLogger.forModule('ReCAPTCHAVerification');
 const { RECAPTCHA_ERROR } = ErrorName;
 
 const verifyRecaptcha = async (recaptchaToken?: string | null) => {
-	const config = getAppConfig();
+	const config = await getAppConfig();
```
mistryrn, line 32: Should we use the static config here since we're only grabbing the recaptcha key?

apps/consent-api/src/routers/user.ts
```diff
@@ -36,40 +37,18 @@ const router = Router();
...
-		firstName: 'Homer',
-		lastName: 'Simpson',
-		email: 'homer.simpson@example.com',
-		role: UserRole.enum.USER,
-		emailVerified: true,
-		enabled2FA: false,
-	};
-	return res.status(200).send({ user });
+router.get('/', withAuth, async (req, res) => {
+	// TODO: return user info - from decoded token?
```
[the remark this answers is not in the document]
mistryrn: That's fair – this is a good candidate for testing out the `withAuth` middleware, plus we have a follow-up ticket to actually implement the `/user` endpoint: https://github.com/OHCRN/platform/issues/167

apps/consent-api/src/server.ts
```diff
@@ -27,7 +27,7 @@ import App from './index.js';
 let server: Server;
 
 (async () => {
-	const appConfig = getAppConfig();
+	const appConfig = await getAppConfig();
```
mistryrn, line 30: Similarly to the previous example – do we need the full app config here, or should we just use the static one? I think the `appConfig` var here gets passed into the `App` function in src/index.js and the only config value it uses is the `express.port`: https://github.com/OHCRN/platform/blob/3bcb2ff664d69c54d25550c34bbaeddb6c1a168a/apps/consent-api/src/index.ts#L50

apps/consent-api/src/services/create.ts
```diff
@@ -64,7 +64,7 @@ type CreateInviteFailureStatus = SystemError | 'INVITE_EXISTS';
 export const createInvite = async (
 	inviteRequest: ClinicianInviteRequest,
 ): Promise<Result<ClinicianInviteResponse, CreateInviteFailureStatus>> => {
-	const { dataMapperUrl } = getAppConfig();
+	const { dataMapperUrl } = await getAppConfig();
```
mistryrn, line 67: If we're just grabbing the `dataMapperUrl`, maybe we can get away with just the static config?

apps/consent-api/src/services/create.ts
```diff
@@ -101,7 +101,7 @@ type CreateParticipantFailureStatus = SystemError | 'PARTICIPANT_EXISTS';
 export const createParticipant = async (
 	request: CreateParticipantRequest,
 ): Promise<Result<CreateParticipantResponse, CreateParticipantFailureStatus>> => {
-	const { dataMapperUrl } = getAppConfig();
+	const { dataMapperUrl } = await getAppConfig();
```
mistryrn, line 104: Same comment as above 🙂

apps/consent-api/src/services/search.ts
```diff
@@ -54,7 +54,7 @@ export const getInformedConsentResponses = async (
 	participantId: string,
 ): Promise<Result<InformedConsentResponse, GetResponsesFailureStatus>> => {
 	try {
-		const { dataMapperUrl } = getAppConfig();
+		const { dataMapperUrl } = await getAppConfig();
```
mistryrn, line 57: Same as above 🙂

apps/consent-api/src/services/search.ts
```diff
@@ -91,7 +91,7 @@ export const getInvite = async (
 	inviteId: string,
 ): Promise<Result<ClinicianInviteResponse, GetInviteFailureStatus>> => {
 	try {
-		const { dataMapperUrl } = getAppConfig();
+		const { dataMapperUrl } = await getAppConfig();
```
mistryrn, line 94: Same as above 🙂

apps/consent-api/tests/config.ts
```diff
@@ -25,6 +25,8 @@ import { AppConfig } from '../src/config.js';
 export const mockEnv = (values?: Partial<AppConfig>) => {
 	// Need to set the NODE_ENV since there are currently places in the code checking the process.env.NODE_ENV directly instead of relying on our config.
 	vi.stubEnv('NODE_ENV', values?.isProduction ? 'production' : 'development');
+	// Need to stub KEYCLOAK_PUBLIC_KEY to prevent the AuthConfig from throwing an error, because there is no external fetch for the public key in testing
```
mistryrn, line 28: Thanks for calling this out 👍

apps/consent-api/tests/routers/invites.spec.ts
```diff
@@ -84,7 +84,7 @@ describe('POST /invites', () => {
 	});
 
 	it('Valid request - makes a POST request to data-mapper and returns created invite with id, inviteAccepted, and inviteSentDate', async () => {
-		const appConfig = getAppConfig();
+		const appConfig = await getAppConfig();
```
mistryrn, line 87: I wonder if we can use the static config in these tests too?

apps/consent-api/tests/routers/participants.spec.ts
```diff
@@ -69,7 +69,7 @@ describe('POST /participants', () => {
 	});
 
 	it('Valid request - makes a POST request to data-mapper and returns created participant with id', async () => {
-		const appConfig = getAppConfig();
+		const appConfig = await getAppConfig();
```
mistryrn, line 72: Same as above 🙂

apps/consent-api/tests/routers/steps/informed-consent.spec.ts
```diff
@@ -50,7 +50,7 @@ describe(`GET ${InformedConsentPath}`, () => {
 	});
 
 	it('Valid request - makes a GET request to data-mapper and returns most recent participant responses for Informed Consent step', async () => {
-		const appConfig = getAppConfig();
+		const appConfig = await getAppConfig();
```
mistryrn, line 53: Same as above 🙂

README.md
```diff
@@ -78,6 +78,8 @@ To run the setup locally, ensure you have provided the **required** environment
...
+| `consent-api` | `KEYCLOAK_REALM_PATH`             | URL for fetching Keycloak public key                                                                                                                                                                                                                                  | `string`  | Optional only if `KEYCLOAK_PUBLIC_KEY` not provided |                                                                                                                                         |
+| `consent-api` | `KEYCLOAK_PUBLIC_KEY`             | Public key for Keycloak JWT parsing                                                                                                                                                                                                                                 | `string`  | Required if `KEYCLOAK_REALM_PATH` not provided |                                                                                                                                         |
```
mistryrn, line 82: I suppose these are both technically optional if the presence of one can make the other optional? 🤔 I reformatted these lines and updated the descriptions to include the "Required if X not provided" parts. What do you think?

```suggestion
| `consent-api` | `KEYCLOAK_REALM_PATH`         | URL for fetching Keycloak public key. Required if `KEYCLOAK_PUBLIC_KEY` is not provided.                                                                                                                                                                | `string`  | Optional |                                                                                                                                                                |
| `consent-api` | `KEYCLOAK_PUBLIC_KEY`         | Public key for Keycloak JWT parsing. Required if `KEYCLOAK_REALM_PATH` is not provided.                                                                                                                                                                 | `string`  | Optional |                                                                                                                                                                |
```

mistryrn (commented):
Looks good! I like the new changes to the middleware and handling the public key fetch. Most of my comments are just places where we can probably use just the "static" config instead of having to call the full async app config function 🙂

apps/consent-api/src/server.ts
anncatton: setting the full AppConfig here so that the auth settings are instantiated on server start

mistryrn: Ah, that makes sense. Thanks!

apps/consent-api/src/authConfig.ts
anncatton: oh wait nope, cuz logger uses staticConfig which is not initialized yet

apps/consent-api/src/middleware/auth.ts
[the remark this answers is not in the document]
anncatton: yeah, i think we can add more in later tasks if we find we need extra data!

apps/consent-api/src/authConfig.ts
mistryrn: Ah, gotcha. Thanks for confirming!

apps/consent-api/tests/config.ts
```diff
@@ -25,7 +25,10 @@ import { AppConfig } from '../src/config.js';
...
+	vi.stubEnv('KEYCLOAK_PUBLIC_KEY', 'mock-public-key');
 
+	// using the full AppConfig here because the tests use a mock App instance
```
mistryrn, line 31: Thanks for calling this out 👍

mistryrn (approved):
This looks good to me! Hopefully Jon can give it a look as well, but I'll approve for now anyway 🤠

apps/consent-api/src/server.ts
joneubank: Maybe leave a comment in the code explaining this ;)

apps/consent-api/tests/config.ts
joneubank: Perhaps creating a stub for the `authConfig.getPublicKey` function would be a more reliable way to ensure that unit tests don't actually fetch things, and allow us to test cases with and without this env property being set ;).

apps/consent-api/tests/config.ts
joneubank: I don't understand this comment.

joneubank (changes requested):
This is all very good and functionally we're basically good to go.

Most of my feedback is minor code organization things, but I am requesting changes cause I would like to simplify the config setup. Specifically:
- Do all the `process.env` access from the `config.ts` file and not from other specialized config files
- keep the main config object static (not async), probably doesnt need a rename
- make the authConfig details accessed through the async getter.
- keep both config objects separate. They can each have their own type.

The other place you should update is the modifications the middleware makes to the request object, I think making the middleware populate an optional object will help organize any future changes to our modification of the shared express request type (changes from this middleware adding more detail, or other middlewares also adding values).

apps/consent-api/src/middleware/recaptcha.ts
```diff
@@ -38,7 +40,7 @@ const verifyRecaptcha = async (recaptchaToken?: string | null) => {
 
 	try {
 		const recaptchaVerification = await axios.post(
-			`https://www.google.com/recaptcha/api/siteverify?secret=${config.recaptcha.secretKey}&response=${recaptchaToken}`,
+			`https://www.google.com/recaptcha/api/siteverify?secret=${secretKey}&response=${recaptchaToken}`,
```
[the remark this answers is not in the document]
anncatton: I'll open a ticket for this so we don't forget

apps/consent-api/tests/config.ts
anncatton: I ended up being able to delete the env stub once the async authConfig was removed from the main AppConfig object, and all the tests can rely on accessing the static config. Unless i'm missing something this should be ok to leave out entirely.

apps/consent-api/src/middleware/auth.ts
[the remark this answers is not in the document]
anncatton: I added the expanded object to the request as you described, with an example `if req.authMiddleware, return success` on the `/user` endpoint. But i'm wondering, with the way the auth function is set up to throw errors if any step in the jwt validation/decoding fails, is it even necessary to have the extra check in the endpoint? I.e. `withAuth` would fail and return a 401 before it even got back to the endpoint, so if we get back to the `/user` handler we can assume the auth check succeeded? (this worked locally, but wondering if i am making a wrong assumption)

apps/consent-api/src/middleware/auth.ts
```diff
@@ -0,0 +1,92 @@
...
+const parseBearerToken = (token?: string): string => {
+	return token && token.startsWith('Bearer ') ? token.replace('Bearer ', '') : '';
+};
+
+/**
+ * Router middleware to ensure the incoming request includes a valid access token in the Authorization header.
+ * When added to a router, it is expected that the incoming request has a header containing a valid access token
+ * in the JWT Bearer format. Any request missing this token or with an invalid token will be rejected with an HTTP 401 status.
+ *
+ * If the token is valid, the "sub" value from the decoded token will be added to the Request object as the "keycloakId"
```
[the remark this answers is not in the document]
anncatton: I added in the `keycloakSubjectId` as the `userId` - but i wasn't entirely sure what you intended that value to be, since you have the entire `verifiedToken` in the obj as well?

joneubank (approved):
Great work here Ann, thanks!

Issue #533, opened 2024-04-02
Update ZOD Schemas to Support New E-Signature Fields


b-f-chan:
Following the creation of the new `Signature` model in Consent DAS to store E-Signatures (https://github.com/OHCRN/platform/issues/532), we need to update any relevant Request/Response Zod schemas to support the new E-Signature fields.

## Schemas to Review

- [x] `types/src/entities/ConsentReviewSign.ts`
- [x] `types/src/services/consentDas/*`
- [ ] `types/src/services/consentApi/*/ConsentReviewSign.ts` (does not currently exist, will likely be added in https://github.com/OHCRN/platform/issues/213 and https://github.com/OHCRN/platform/issues/212)
- [x] `types/src/services/consentUi/*/ConsentReviewSign.ts` (does not currently exist, this ticket could be an opportunity to add it)

## Changes to Make

- [x] (if not already covered in https://github.com/OHCRN/platform/issues/532) create Signature entity base type
    - [x] id should be a NanoId
- [x] ensure types for requests that submit e-signatures contain the necessary new signature info:
    - [x] signature
    - [x] createdAt
    - [x] authenticationMethod (for now we will just use the dummy string `'multi-factor auth'`, but in a future update we may change this to a field for the session ID of the user creating the signature)
- [x] ensure types for responses that return e-signatures return the necessary signature info:
    - [x] signature
    - [x] createdAt

Issue #532, opened 2024-04-02
Consent DAS: Add E-Signature Model to Prisma Schema


b-f-chan:
Add a new `Signature` model to the Consent DAS Prisma schema to capture e-signature fields, and a new router endpoint to the Consent DAS for storing and retrieving e-signatures.

**Note: the work in this ticket should be limited to the Consent DAS and the Types library.**

## `schema.prisma` Updates

The model should be called `Signature` and contain the following fields:

- id (String, required, default `""`, `@db.Char(21)`)
- participantId (String)
- signature (String, required)
- createdAt (DateTime, required, default `now()`)
- authenticationMethod (String, required)

Similar to the `ParticipantResponse` model in Consent DAS, it should reference the `id` of `Participant` as a foreign key, `participantId`. You will need to add the appropriate links between the two models (reference [ParticipantResponse](https://github.com/OHCRN/platform/blob/a54ea04c2bb16606d2466016cde301e390b4f5c0/apps/consent-das/prisma/schema.prisma#L73) as an example).

## `prismaClient.ts` Updates

The Prisma Client ([src/prismaClient.ts](https://github.com/OHCRN/platform/blob/develop/apps/consent-das/src/prismaClient.ts)) will need to be updated to use `nanoid()` for the IDs of this new model, similar to this example for [participantResponse](https://github.com/OHCRN/platform/blob/a54ea04c2bb16606d2466016cde301e390b4f5c0/apps/consent-das/src/prismaClient.ts#L79).

## Router Updates

Once the model is added to the Prisma schema, we'll need to add a Consent DAS `/signatures` router endpoint for handling GET and POST operations for `Signature`s. Follow the [src/routers/participantResponses.ts](https://github.com/OHCRN/platform/blob/develop/apps/consent-das/src/routers/participantResponses.ts) example and add a POST `/` endpoint, and a GET `/:participantId/:signatureId` endpoint.

This will require creating the necessary functions in `src/services/search.ts` and `src/services/create.ts` to support the GET and POST endpoints (for example, something like `getSignatures`, `getSignatureById`, `createSignature`, etc – these are not exact functions/names, just examples. Please refer to the existing participant response examples), as well as any corresponding types in the shared types library.

## Checklist
- [x] add `Signature` model to Consent DAS Prisma schema
- [x] connect `Signature` model to `Participant` model, using `ParticipantResponse` as a reference
- [x] update the `prismaClient` to use `nanoid()` for IDs on the new `Signature` model
[...]

PR #537, opened 2024-04-04
🗃️#532: Consent DAS: Add E-Signature Model to Prisma Schema

xusandrew:
## Description of Changes

- Adds signature model to consent DAS prisma schema
- Adds `getSignatureById` and `createSignature` methods
- `/signatures` router with a POST at `/` and a GET at `/:participantId/:signatureId`
- Created some necessary req/res types in the types package, but will update in #533
- Tested using swagger/pgadmin

### Special Instructions
Before running these changes, you will need to build
```
pnpm build
```
also run migrations
```
pnpm prisma:migrate:dev
```

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-das/prisma/schema.prisma
```diff
@@ -94,3 +95,12 @@ model ClinicianInvite {
   inviteAcceptedDate                 DateTime?    @db.Date
   inviteAccepted                     Boolean      @default(false)
 }
+
+model Signature {
+  id                   String       @id @default("") @db.Char(21)
+  participant          Participant? @relation(fields: [participantId], references: [id], onDelete: Restrict, onUpdate: Restrict)
+  participantId        String?
```
mistryrn, line 102: Apologies, I just noticed I had this in the ticket as just `participantId (String)`, but it should be required as well. Do you mind updating this to be required?

packages/types/src/services/consentDas/requests/Signature.ts
```diff
@@ -0,0 +1,31 @@
...
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+import { z } from 'zod';
+
+import { NanoId } from '../../../entities/fields/index.js';
+
+export const SignatureRequest = z.object({
+	participantId: NanoId,
+	signatureId: NanoId,
+});
```
mistryrn, line 29: We'll need a request object schema for the `POST` request as well, containing:
- [ ] participantId
- [ ] signature
- [ ] authenticationMethod
- [ ] createdAt

mistryrn (changes requested):
Really good stuff in here, just a few things missing because my requirements/examples didn't contain everything!

Requesting a few small changes here, mainly just to reference the Clinician Invite flow as an example since we had implemented our full validation and error handling by that point.

apps/consent-das/src/services/create.ts
```diff
@@ -145,3 +146,49 @@ export const createClinicianInvite = async (
...
+ * @param authenticationMethod Authentication method used to create the signature
+ * @returns Signature object from Consent DB
+ */
+export const createSignature = async ({
+	participantId,
+	signature,
+	authenticationMethod,
+	createdAt,
+}: {
+	participantId?: string;
```
mistryrn, line 164: Let's make `participantId` required here and update the JSDoc comment above 🙂 
```suggestion
	participantId: string;
```

mistryrn (commented):
Thanks for getting these changes in so quickly! It's looking good, just a couple more small changes and we should be good to go with this 🙏

anncatton (changes requested):
Lots of good stuff here Andrew! 

The changes for the migrations and the get + create behaviour i'd like to see done in this PR, plus adding the base type for Signature.

For the other type stuff I'm good with that being in #533 , but i'd like to add `TODO`s wherever we're postponing work so it's very clear what's still to be done.

apps/consent-das/prisma/migrations/20240404135907_add_signature_model/migration.sql
```diff
@@ -0,0 +1,13 @@
+-- CreateTable
+CREATE TABLE "Signature" (
```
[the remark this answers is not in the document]
xusandrew: very weird i saw this too!

apps/consent-das/prisma/schema.prisma
```diff
@@ -94,3 +95,12 @@ model ClinicianInvite {
...
   inviteAccepted                     Boolean      @default(false)
 }
+
+model Signature {
+  id                   String      @id @default("") @db.Char(21)
+  participant          Participant @relation(fields: [participantId], references: [id], onDelete: Restrict, onUpdate: Restrict)
+  participantId        String
+  signature            String
+  createdAt            DateTime    @default(now())
+  authenticationMethod String
```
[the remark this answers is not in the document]
mistryrn: Good catch! Yes, let's add a composite ID constraint here too 🙂

apps/consent-das/src/services/search.ts
```diff
@@ -182,3 +183,50 @@ export const getClinicianInvites = async (): Promise<ClinicianInvite[]> => {
...
+ */
+export const getSignatureById = async ({
+	signatureId,
+	participantId,
+}: {
+	signatureId: string;
+	participantId: string;
+}): Promise<Result<Signature, GetSignatureFailureStatus>> => {
+	// Check that participant exists, return failure if not
+	const participantResult = await getParticipantById(participantId);
```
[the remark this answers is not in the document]
mistryrn: Yeah, my bad – that's on me. I was referencing `participantResponses` too heavily while creating the ticket when the behaviour for the `/signatures` router is different.

The Consent DAS endpoints we’ll need for the `/signatures` router are:
- `POST /`  — create signature
- `GET /:participantId` — get all signatures for participant, sorted by `createdAt` date

Based on this, the search function should return an array of signatures objects for a given participant and could probably be called `getParticipantSignatures` instead to make things clear.

At the Data-Mapper layer we will then pick the latest signature (if any) from the array, and return that to the Consent-API, but that's not part of this ticket.

mistryrn (commented):
Just following-up on Ann's review, let's discuss more in person when we're all in office

apps/consent-das/src/services/create.ts
```diff
@@ -145,3 +150,43 @@ export const createClinicianInvite = async (
...
+		.then((signature) => success(signature))
+		.catch((error) => {
+			if (error instanceof PrismaClientKnownRequestError) {
+				if (error.code === 'P2002') {
+					const errorMessage = `A signature already exists with that '${
+						error.meta?.target ?? 'data'
+					}'`;
+					logger.error('POST /signatures', errorMessage, error.message);
+					return failure('SIGNATURE_EXISTS', errorMessage);
+				}
```
anncatton, line 178: i ran into this just by using the default `participantId` in swagger when testing locally - can you add a block to handle an error on the participantId? prisma will throw foreign key constraint error (code `P2003`), which just means the `participantId` can't be found in the consent DAS `Participant` table. This error can be `PARTICIPANT_DOES_NOT_EXIST`, and you can update the `CreateSignatureFailureStatus` type to include that. Then just add in that case on the endpoint like you've done with `SIGNATURE_EXISTS`.

This is somewhat similar to what you were trying out earlier with the query for the participantId in the Participant table before the create query, but this way we can take advantage of the errors Prisma is giving us! We don't actually have this in the participant Response flow but it may be worth adding in a separate task.

I was looking up what HTTP error makes sense here and it *seems* like a 409 might be appropriate? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409 (my first intuition was 400 but i don't think that's right). what do you and @mistryrn think?

packages/types/src/services/consentDas/responses/Signature.ts
```diff
@@ -0,0 +1,37 @@
...
+import { generateSchema } from '@anatine/zod-openapi';
+import type { SchemaObject } from 'openapi3-ts/oas31';
+import { z } from 'zod';
+
+import { NanoId } from '../../../entities/fields/index.js';
+import { SignatureBase } from '../../../entities/index.js';
+
+export const SignatureResponse = SignatureBase.extend({
+	signatureId: NanoId,
+	signature: z.string(),
```
anncatton, line 29: See my above comment about extending the base with an additional subtype.

anncatton (commented):
Nice work! I've added some more suggestions for error handling and a bit of refactoring

apps/consent-das/src/routers/signatures.ts
```diff
@@ -0,0 +1,170 @@
...
+			const signature = await createSignature(req.body);
+			switch (signature.status) {
+				case 'SUCCESS': {
+					return res.status(201).json(signature.data);
+				}
+				case 'SIGNATURE_EXISTS': {
+					return res.status(409).json(ConflictErrorResponse(signature.message));
+				}
+				case 'PARTICIPANT_DOES_NOT_EXIST': {
+					return res.status(404).json(NotFoundErrorResponse(signature.message));
```
anncatton, line 157: i chatted with @mistryrn on this, let's make this a 400 error. The error message can stay the same.

anncatton (commented):
looking great Andrew! Just 2 smol changes needed.

anncatton (approved):
Great work Andrew! 🚀

Issue #535, opened 2024-04-04
Consent UI: Make enum for error codes


samrichca:
We're re-using error code strings in Zod refine/superrefine, specifically for UI validation.

- [ ] Add strings like `guardianHasParticipantInfo` to an enum
- [ ] Update instances of these strings to use the enum

PR #548, opened 2024-04-10
🏷️ #535: make an enum for zod custom errors

samrichca:
### Types
- Add an enum with the Zod custom errors
- Update types and tests

## PR Readiness Checklist

- [x] Connected ticket to PR


packages/types/src/services/consentUi/utils/zodCustomErrors.ts
```diff
@@ -0,0 +1,35 @@
...
+	'guardianInfoMissing',
+	'participantEmailMissing',
+	'participantHasGuardianEmail',
+	'participantHasGuardianInfo',
+	'participantLessThanMinimumAge',
+	'participantMissingPhoneNumber',
+] as const;
+
+export const ZodCustomError = z.enum(ZOD_CUSTOM_ERRORS);
+export type ZodCustomError = z.infer<typeof ZodCustomError>;
```
mistryrn, line 35: How would you feel about trying Zod's `z.nativeEnum` here: https://zod.dev/?id=native-enums

This would allow us to make the `ZOD_CUSTOM_ERRORS` an object with keys, so instead of using `ZodCustomError.enum.participantMissingPhoneNumber`, we can use `ZOD_CUSTOM_ERRORS.PARTICIPANT_MISSING_PHONE_NUMBER` when we want to reference a specific custom error.

```suggestion
export const ZOD_CUSTOM_ERRORS = [
	GUARDIAN_EMAIL_MISSING: 'guardianEmailMissing',
	GUARDIAN_HAS_PARICIPANT_EMAIL: 'guardianHasParticipantEmail',
	GUARDIAN_HAS_PARTICIPANT_PHONE_NUMBER: 'guardianHasParticipantPhoneNumber',
	GUARDIAN_INFO_MISSING: 'guardianInfoMissing',
	PARTICIPANT_EMAIL_MISSING: 'participantEmailMissing',
	PARTICIPANT_HAS_GUARDIAN_EMAIL: 'participantHasGuardianEmail',
	PARTICIPANT_HAS_GUARDIAN_INFO: 'participantHasGuardianInfo',
	PARTICIPANT_LESS_THAN_MINIMUM_AGE: 'participantLessThanMinimumAge',
	PARTICIPANT_MISSING_PHONE_NUMBER: 'participantMissingPhoneNumber',
} as const;

export const ZodCustomError = z.nativeEnum(ZOD_CUSTOM_ERRORS);
export type ZodCustomError = z.infer<typeof ZodCustomError>;
```

mistryrn: After discussing with @anncatton and as per the Zod docs recommendation, you can ignore this suggestion and continue with the regular Zod `enum` instead of `nativeEnum`!

packages/types/src/services/consentUi/utils/zodCustomErrors.ts
```diff
@@ -0,0 +1,35 @@
+/*
```
anncatton, line 1: Let's rename this file to match the type convention we have(and your actual type here 🙂 )  - `ZodCustomError`, singular.

packages/types/src/services/consentUi/utils/guardianFields.ts
```diff
@@ -119,13 +119,13 @@ export const hasRequiredGuardianInfoForRegistration = (
 	if (isGuardian) {
 		Object.entries(fields).forEach(([key, value]) => {
 			if (isEmptyOrUndefined(value?.trim())) {
-				addZodCustomError(ctx, key, 'guardianInfoMissing');
+				addZodCustomError(ctx, key, ZodCustomError.enum.guardianInfoMissing);
```
anncatton, line 122: Using the enum type as the message is correct - but looks like this still needs to be done for the errors returned from `hasRequiredContactInfoForInvite`.

anncatton (commented):
This is great, thanks for getting the ball rolling on this!

A couple suggestions:
- we should type the `addZodCustomError` `message` param as a `ZodCustomError`. Then, in the guardian/participant refinements, vscode will show where the error messages are not correctly typed, and should enforce using a standard enum for any newly added errors
- for readability, let's use the destructuring approach we have for other enums:
```
const { guardianInfoMissing } = ZodCustomError.enum;
```

samrichca:
closing due to filename casing errors

PR #551, opened 2024-04-12
🏷️ 485 - supplementary - correct response types for register form

anncatton:
## Description of Changes

Changes type of `consentToBeContacted` field in invite flow responses. `z.boolean()` did not work with the `z.literal(true)` in the Request types, because it *can* be `false`

**Note:** Similar changes _may_ be required for the `CreateParticipant` types, since the same strategy of `z.literal(true)` in the request and `z.boolean()`, depending on form behaviour.

### Consent UI
- because we're using the `Result<Success, Failure>` type for the response of `handleGetInvite()`, a `failure()` response will return void. So we set the result as `undefined` in this case to match what the `RegisterFormWrapper` is expecting. We should investigate nicer response handling here
- Adds some error logging for testing purposes, until error handling on the UI is implemented

### Types
- changes type of `consentToBeContacted` in `ClinicianInvite` response schemas to `z.literal(true).optional()` to work with register form, when invite data is inserted into the form

## PR Readiness Checklist

- [ ] Connected ticket to PR


samrichca (approved):
this looks great, thanks

Issue #554, opened 2024-04-12
Invite page: Remove assent form identifier, add checkbox


samrichca:
- Collecting the assent form will now be a business process rather than a form field.
- There needs to be a checkbox for confirming the user knows they need to send an assent form.
- Design & verbiage TBA
- This ticket only impacts consent API and consent UI.

To do:
- [ ] Remove assent form identifier from consent API invite request, refinements, and tests.
- [ ] Add a checkbox field, for the "guardian of minor with assent" consent group, in the UI
- [ ] Make this checkbox field conditionally required for the consent group in Zod schemas
  - [ ] Add tests

---

from #556 

Data Mapper and DAS changes required to support the new Assent Form business process:

**Prisma:**

- [x] Consent DAS: on the ClinicianInvite and Participant models in the schema.prisma, add an optional Boolean called assentFormAcknowledged
- [x] Consent DAS: update clinician invite router to support passing and returning assentFormAcknowledged in the request/response (will require updating types/consentDas/*/ConsentClinicianInvite.ts Zod types as per details in Zod section below)

**Zod:**

- [x] Types: remove assentFormIdentifier from the relevant Clinician Invite (PIClinicianInvite) request and response types in types/services/piDas and types/services/dataMapper)
- [x] Types: add assentFormAcknowledged to the relevant Clinician Invite (ClinicianInvite and ConsentClinicianInvite) request and response types (in types/services/consentDas and types/services/dataMapper )
- [x] Types: rename hasRequiredAssentFormIdentifier util to hasRequiredAssentFormAcknowledgement, update the function to check that assentFormAcknowledged is true
- [x] Types: remove usage of old hasRequiredAssentFormIdentifier util on relevant types/services/piDas types
- [ ] Types: replace usage of old hasRequiredAssentFormIdentifier util on relevant types/services/dataMapper types with hasRequiredAssentFormAcknowledgement
- [x] Types: add hasRequiredAssentFormAcknowledgement util to relevant types/services/consentDas types

**Notes:**

- We can leave the existing Prisma schema fields for assentFormIdentifier in place until we work on the CRC Dashboard and move them to their appropriate place
- Consent API Zod type changes should be similar to the Data Mapper ones

samrichca:
invite PR:
- add assetFormAcknowledged to prisma schema for ClinicianInvite only

PR #565, opened 2024-04-22
🏷️ 🗃️ #554 1/3 - add assentFormAcknowledged to invite backend

samrichca:
## Description of Changes

only covers the invite flow in the backend.

### Consent DAS
- add assentFormAcknowledged to ClinicianInvite table
- update swagger

### Data Mapper
- swap assentFormIdentifier for assentFormAcknowledged when creating an invite

### Types
- add hasRequiredAssentFormAcknowledgement refinement
- add reusable assentFormAcknowledgement field schemas for requests & responses

### Consent API
- add assentFormAcknowledged to ClinicianInvite & ConsentClinicianInvite

## PR Readiness Checklist

- [x] Connected ticket to PR


packages/types/tests/utils/conditionalFields.spec.ts
```diff
@@ -274,4 +275,60 @@ describe('Conditional fields utility functions', () => {
 			expect(result).false;
 		});
 	});
+
+	describe('hasRequiredAssentFormAcknowledgement', () => {
+		it('returns TRUE if consent group includes assent and assentFormIdentifier is true', () => {
```
samrichca, line 280: i see this but i don't want to update the PR while it's in review

anncatton: lol you read my mind...i saw this yesterday, thanks for leaving a comment so it's not missed

packages/types/tests/entities/ClinicianInvite.spec.ts
```diff
@@ -210,40 +225,54 @@ describe('ClinicianInviteRequest', () => {
 });
 
 describe('ConsentClinicianInviteResponse', () => {
+	const testObj = {
```
anncatton, line 228: nice! thanks for tidying

packages/types/src/services/consentDas/responses/ConsentClinicianInvite.ts
```diff
@@ -22,13 +22,15 @@ import { generateSchema } from '@anatine/zod-openapi';
...
 import {
+	AssentFormAcknowledgedNullableField,
 	ConsentGroupField,
 	InviteClinicianFields,
 	InviteEntity,
 } from '../../../entities/ClinicianInvite.js';
 
 export const ConsentClinicianInviteResponse = InviteEntity.merge(InviteClinicianFields)
 	.merge(ConsentGroupField)
+	.merge(AssentFormAcknowledgedNullableField)
```
anncatton, line 33: This works so not asking for any changes, but thinking I need to revisit where we're calling
this won't break anything so you can leave as is, but we shouldn't need the transform here because we have it on the data-mapper das type, which performs the transform when we call `safeParse()` on the response in data-mapper das service. It should work as just `z.literal(true).nullable()`

packages/types/src/services/dataMapper/requests/ClinicianInvite.ts
```diff
@@ -23,17 +23,21 @@ import { z } from 'zod';
...
+} from '../../../entities/ClinicianInvite.js';
 
 export const ClinicianInviteRequest = ClinicianInviteBase.merge(ConsentToBeContacted)
+	.merge(AssentFormAcknowledgedRequestField)
 	.refine(hasRequiredInfoForConsentGroup, {
 		message: 'Contact fields must be related to consentGroup',
 	})
-	.refine(hasRequiredAssentFormIdentifier, {
-		message: 'Invalid assentForm data provided for consentGroup',
+	.refine(hasRequiredAssentFormAcknowledgement, {
```
anncatton, line 39: 👍

packages/types/src/services/dataMapper/responses/ConsentClinicianInvite.ts
```diff
@@ -22,13 +22,15 @@ import { generateSchema } from '@anatine/zod-openapi';
...
 import {
+	AssentFormAcknowledgedNullableField,
 	ConsentGroupField,
 	InviteClinicianFields,
 	InviteEntity,
 } from '../../../entities/ClinicianInvite.js';
 
 export const ConsentClinicianInviteResponse = InviteEntity.merge(InviteClinicianFields)
 	.merge(ConsentGroupField)
+	.merge(AssentFormAcknowledgedNullableField)
```
anncatton, line 33: as mentioned above, i think it's better to define the transform here, inside the `extend()`, for readability, and it's not being reused.

anncatton (commented):
Great work @samrichca , this was very clean and easy to review. I don't see any major issues, the comments I've added are mostly "i think this would be a nicer implementation". Plus it's been a little bit since the invite types were worked on in a big way, and I've changed my mind a little/learned a bit so some of the comments reflect that as well. Let me know if some of these suggestions don't work or cause problems, I played with a bit locally but did not try every create scenario.

Also don't forget about the test description typo you mentioned!

apps/consent-das/src/routers/swagger-schemas/clinician-invite.yaml
```diff
@@ -5,20 +5,20 @@ components:
       properties:
         id:
           type: string
-          example: "clmarsvhd000008jngksv"
+          example: 'clmarsvhd000008jngksv'
```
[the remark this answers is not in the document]
anncatton: i mean, to follow the example of how the Request schema is imported. You already have the corresponding Response type, you just need to import it so it's available for swagger.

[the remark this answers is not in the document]
samrichca: this created errors so I think we should shelve it for now

image

packages/types/src/services/consentDas/requests/ConsentClinicianInvite.ts
```diff
@@ -29,7 +29,8 @@ export const ConsentClinicianInviteRequest = z
 	.object({ id: NanoId })
 	.merge(InviteClinicianFields)
 	.merge(ConsentToBeContacted)
-	.merge(ConsentGroupField);
+	.merge(ConsentGroupField)
+	.merge(z.object({ assentFormAcknowledged: z.literal(true).optional() }));
```
anncatton, line 33: just for extra safety, let's add the assent check here as well! In the full flow of course the check occurs earlier, but it's nice to have another verification at each service layer.

apps/consent-das/src/routers/swagger-schemas/clinician-invite.yaml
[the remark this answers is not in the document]
anncatton: Can you double check there's not some weird import from `src` or something like that? I have verified this works on my local, even with the existing yaml file in the swagger config

anncatton (commented):
Thanks for simplifying the optional literal fields, this is much more readable. Just a few more comments added.

Also agree that the nullable/transform stuff is good as is.

samrichca:
@anncatton I updated the branch from develop and now this change works, I'm not sure why, but it does 👍

packages/types/src/services/consentDas/requests/ConsentClinicianInvite.ts
samrichca: should it be in data mapper too?

anncatton: i thought about this but i think no, i was more concerned about the types where the data is first consumed by a service...it's unlikely that the underlying das service call is going to be accessed directly, outside of the data-mapper create call

apps/consent-das/src/routers/swagger.ts
```diff
@@ -31,6 +31,7 @@ import {
 } from 'types/entities';
 import {
 	ConsentClinicianInviteRequestSchema as ClinicianInviteRequest,
+	ConsentClinicianInviteResponseSchema as ClinicianInviteResponse,
```
anncatton, line 34: 👍

anncatton (approved):
Thanks @samrichca 🚀

PR #581, opened 2024-05-05
✨ 🏷️ 258 - data mapper step 2 POST

anncatton:
## Description of Changes
Step 1) of the changes needed for the `POST` endpoint for the consentReleaseData. Covers PI and Consent operations at the Data Mapper level, and outlines the approach for the full set of steps needed for each DAS service

### Data Mapper
- adds `consentReleaseData` router
- adds `POST/{participantId}` endpoint
- adds initial version of `createConsentReleaseData` in the `services/create` dir, with jsdocs to describe the steps involved
- in `createConsentReleaseData`, adds steps to update participant PI data and create participant consent responses for `RELEASE_DATA__DE_IDENTIFIED` and `RELEASE_DATA__CLINICAL_AND_GENETIC` questions. Underlying consent das functionality already implemented in https://github.com/OHCRN/platform/issues/256
- adds `updateParticipantPiData` service call; underlying PI DAS functionality implemented in https://github.com/OHCRN/platform/pull/582

### Types
- Updates `ConsentReleaseData` entity type to match `ohipName` convention, + adds `selfReportedPrimaryCancerDiagnosis`
- Adds `ConsentReleaseData` and `PIUpdateParticipant` req/res types for Data Mapper
- Adds refinement functions for conditional `selfIdentifiedGender` and `selfReportedPrimaryCancerDiagnosis` fields + adds unit tests
- adds some `TODO` comments in `ConsentReviewSign` and `ConsentReleaseDataBase` entity types regarding stop gap fix for tests/outstanding issues.

### Docs
- adds a sequence diagram to describe all the steps needed for the ConsentReleaseData POST operation

## PR Readiness Checklist

- [ ] "Expected Outcome(s)" in ticket have been met
- [x] Ticket number included in PR title
- [x] Connected ticket to PR
- [x] Labels added to PR for service name (`consent-api`, `data-mapper`, etc...), type (`chore`, `documentation`, etc...), status (`draft`, `on-hold`, etc...) **if applicable**
- [ ] Manual testing completed
- [ ] Builds locally without errors or warnings
- [x] Tests are updated (if required) and passing
- [ ] PR feedback has been addressed **if applicable**
- [ ] New environment variables added to `.env.schema` files, `README.md`
- [ ] Added copyrights to new files

PR #585, opened 2024-05-06
✨ 🏷️ 258 - PHI DAS operations

anncatton:
## Description of Changes

Implements PHI DAS operations to support ConsentDataRelease step in Wizard. 

Note: I did not remove any of the old router/service functions that was initially meant to handle these operations. I would prefer to do this as a cleanup task in a followup ticket, as this PR is already rather large!

### Data Mapper
- extracts the operations for the consent question responses and the OhipNumber flow into their own handler functions, just to tidy up `createConsentReleaseDataRequest` a bit
- corrects the return type for `createConsentReleaseDataRequest` and updates the jsdoc to match the new Keys/PHI DAS order of operations: all OhipNumber operations for Keys and PHI will be completed before moving on the operations for ClinicalProfile in Keys and PHI
- adds service calls for PHI requests for OhipNumber and ClinicalProfile
- adds missing jsdocs
- updates some of the service function names to better represent what they're doing: functions meant to perform upsert actions (findOrUpdate, updateOrCreate) are renamed so

### PHI DAS
- adds `upsert` queries to prisma client to support the `updateOrCreate` functions for OhipNumber and ClinicalProfile.
- adds router endpoints for modifying OhipNumber and ClinicalProfile records. Both use `PUT` methods because they will create or update the specific resource.
- updates jsdocs and swagger schemas
- adds service functions for the `updateOrCreate` operations. The prisma queries will use `upsert` to update a record if it is found by private key (`update` query clause), or will create a new record if none exists (`create` query clause).
- adds `express-request-validation` package for validating the new router endpoint requests
 
### Consent UI
- updates some field names in the ConsentReleaseData form to match the updated type

### Types
- adds supporting types for new PHI services and routers
- updates `hasRequiredSelfIdentifiedGender` args to reflect field name changes
- adds a `refinementErrorMessages` dictionary to hold string messages for our conditional refinements. To reduce changes in this PR, only the applicable refinements are addressed.
- refactors `ConsentReleaseData` entity type to use fields from `ClinicalProfile`. This will better align with the prisma schema.
- updates tests that were failing due to the field name changes

### General
- updates the `consent-release-data-flow` diagram to match the change in the order of Keys/PHI DAS ops for Ohip and ClinicalProfile

### Special Instructions
Before running these changes, you will need to install `express-request-validation` for PHI DAS:
```
pnpm i
```

## PR Readiness Checklist

- [ ] Connected ticket to PR


packages/types/src/common/conditionalFieldUtils.ts
```diff
@@ -208,12 +208,12 @@ export const hasRequiredSelfReportedPrimaryCancerDiagnosis = (props: {
  * @returns {boolean} returns true if selfIdentifiedGender condition is met
  */
 export const hasRequiredSelfIdentifiedGender = (props: {
-	genderIdentity: Gender;
+	gender: Gender;
```
anncatton, line 211: update fields to match the changes in `ConsentReleaseData` entity type

packages/types/src/common/refinementErrorMessages.ts
```diff
@@ -0,0 +1,32 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * Dictionary of error messages for conditional field refinements.
+ * For backend error messaging only.
+ */
+export const refinementErrorMessages = {
```
anncatton, line 26: Added this dictionary just because it was getting annoying to add in the error strings for every refinement message.

I've only added the messages relevant to this ticket, we can add the rest + update the other refinement messages in a later task.

packages/types/src/entities/ClinicalProfile.ts
```diff
@@ -28,34 +28,32 @@ import {
...
+	historyOfCancer: HistoryOfCancer,
+	familyHistoryOfCancer: HistoryOfCancer,
+	gender: Gender,
+	selfIdentifiedGender: OptionalString, // only required if genderIdentity = PREFER_TO_SELF_IDENTIFY
+	selfReportedPrimaryCancerDiagnosis: PrimaryCancerDiagnosis.optional(), // only required if historyOfCancer = YES
+	vitalStatus: VitalStatus.optional(),
+});
+export type ParticipantClinicalFields = z.infer<typeof ParticipantClinicalFields>;
 
-		const requiresSelfIdentifiedGender = input.gender === Gender.enum.PREFER_TO_SELF_IDENTIFY;
```
anncatton, line 57: moves the refinement to an external util function to match the approach we've used for the other conditional fields

packages/types/tests/entities/ClinicalProfile.spec.ts
```diff
@@ -27,12 +27,12 @@ import {
 	HistoryOfCancer,
 	VitalStatus,
 } from '../../src/entities/fields/index.js';
-import { ClinicalProfile } from '../../src/entities/index.js';
+import { CreateClinicalProfileRequest } from '../../src/services/dataMapper/requests/CreateClinicalProfile.js';
```
anncatton, line 30: i removed the refinements from the base entity type so this test started to fail. Importing the data-mapper request type instead to account for that change.

packages/types/tests/utils/conditionalFields.spec.ts
```diff
@@ -393,36 +393,36 @@ describe('Conditional fields utility functions', () => {
 	describe('hasRequiredSelfIdentifiedGender', () => {
 
 		const { PREFER_TO_SELF_IDENTIFY, MAN, TRANSGENDER_MAN_TRANSMAN} = Gender.enum
-		it('returns true if genderIdentity is PREFER_TO_SELF_IDENTIFY and selfIdentifiedGender is present', () => {
+		it('returns true if gender is PREFER_TO_SELF_IDENTIFY and selfIdentifiedGender is present', () => {
```
anncatton, line 396: updating tests to match the field name changes in ConsentReleaseData types

packages/types/src/services/piDas/responses/PIUpdateParticipant.ts
```diff
@@ -20,9 +20,13 @@
 import { z } from 'zod';
 import { generateSchema } from '@anatine/zod-openapi';
 
-import { PIParticipantBase } from '../../../entities/index.js';
+import { NanoId, PIParticipantBase } from '../../../entities/index.js';
 
-export const PIUpdateParticipantResponse = PIParticipantBase;
+export const PIUpdateParticipantResponse = PIParticipantBase.merge(
```
anncatton, line 25: Missed the `id` field that should be returned in the response type.

packages/types/src/services/phiDas/responses/CreateClinicalProfile.ts
```diff
@@ -0,0 +1,50 @@
...
+	PrimaryCancerDiagnosis,
+	VitalStatus,
+} from '../../../entities/fields/index.js';
+
+export const CreateClinicalProfileResponse = ClinicalProfile.merge(
+	z.object({
+		clinicalProfilePrivateKey: NanoId,
+	}),
+).extend({
+	selfIdentifiedGender: OptionalString.nullable(),
```
anncatton, line 39: I know we had some discussion on how to handle the null values that come from the DAS/DB layers - updating the prisma request to convert from null to undefined so that none of the TS code needs to deal with this step. For now we can leave what we have in place and make those changes in a followup task (if we decide to go that route).
This flow will use the transform step at the data-mapper DAS service level, as we've done in the other flows.

apps/phi-das/src/routers/ohip.ts
```diff
@@ -94,44 +99,70 @@ router.get('/:ohipPrivateKey', async (req, res) => {
 
 /**
  * @openapi
- * /ohip:
- *   post:
+ * /ohip/{ohipPrivateKey}:
+ *   put:
```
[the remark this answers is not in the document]
mistryrn: Agreed, let's stick with `PUT` here.

packages/types/src/common/refinementErrorMessages.ts
mistryrn: Thanks for this!

mistryrn (commented):
👏 So great to be able to test this! Thanks for all the hard work putting this together.

I've left one renaming suggestion here, but other than that I think it's just my suggestions from #581 that you might want to consider here 🙂

mistryrn (approved):
Thanks for getting these changes in! 😄

Issue #260, opened 2023-11-16
Build Data Mapper Endpoint --> Consent Wizard - POST - Step 3


mistryrn:
Build corresponding `POST` endpoint in the `data-mapper` to consume the validated Consent to Research Participation Request coming from `consent-api` and send it to the relevant DASes to store the Consent to Research Participation.

## DAS Breakdown
- Consent DAS

## Data Mapper Breakdown
- [ ] TBD

## Order of Operations
1. TBD

## Success Response
TBD - POST vs PATCH considerations

## Other Considerations
Highlight areas that may need the following:
- Logging
- Error Handling
- Validation

PR #589, opened 2024-05-07
✨ 🏷️ #260: data mapper step 3 POST

samrichca:
### Data Mapper
- add a POST endpoint for submitting consent for research participation
- tested in swagger & postico

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/data-mapper/src/services/create.ts
```diff
@@ -280,3 +305,77 @@ export const createInformedConsent = async (
...
+		}
+
+		const consentResearchParticipation = {
+			[parsedContactInformationResponse.data.consentQuestionId]:
+				parsedContactInformationResponse.data.response,
+			[parsedFutureResearchResponse.data.consentQuestionId]:
+				parsedFutureResearchResponse.data.response,
+		};
+
+		return success(consentResearchParticipation);
```
mistryrn, line 373: As mentioned in my comment on the response type, here's an example of validating the response type before returning from the create function. We don't **have** to do this, but we can check with Ann in office tomorrow and see if we should.

```suggestion
		const parsedConsentResearchParticipation = ConsentResearchParticipationResponse.safeParse(
			consentResearchParticipation,
		);
		if (!parsedConsentResearchParticipation.success) {
			logger.error(
				'Received invalid data in create Consent for Research Participation response',
				parsedConsentResearchParticipation.error.issues,
			);
			return failure('SYSTEM_ERROR', parsedConsentResearchParticipation.error.message);
		}

		return success(parsedConsentResearchParticipation.data);
```

mistryrn (changes requested):
This is great 🎉  Tested locally and it works well! Just one change request to the response type, plus a suggestion that we can discuss with Ann tomorrow 🙂

apps/data-mapper/src/services/create.ts
mistryrn: Discussed with @anncatton in office, we can proceed with these suggestions since we're combining separate response types into a single response, just to be safe that we're returning what we should 🙂

mistryrn (approved):
Looks good! 👍

PR #701, opened 2024-06-07
🐛 630 - Ohip conflict handling

anncatton:
## Description of Changes

Adds error and rollback handling for conflict errors in the PHI DAS Ohip table. Ohip key rollback is only intended for cases where a participant does not have an existing Ohip number record.

Tested with these scenarios:
1) new user, setting new unique ohip number - ohip key + ohip number record is created, `hasOhip: null` -> `hasOhip: true`
2) same user, updating with unique ohip number - ohip key is preserved + ohip number record is updated, `hasOhip` remains `true`
3) new user, setting non-unique ohip number - ohip record should not be created, ohip key should be rolled back, `hasOhip` remains `null`, error should be returned
4) existing user with unique ohip number, setting non-unique ohip number - should receive error, ohip is not updated, ohip key is not deleted, `hasOhip` remains `true`
5) new user, no ohip number  - `hasOhip: null` -> `hasOhip: false`

### Data Mapper
- switches order of steps in `createConsentReleaseData` to better facilitate Ohip key rollback (thanks for the suggestion @mistryrn !). Keys + PHI DAS operations will be attempted first. If any part of the Ohip op fails, the function will return an error. This prevents the PI `hasOhip` status from being updated prematurely, which was causing a parsing error on the GET for Step 2 - `hasOhip` cannot be true if there is no Ohip record related to the Participant.
- updates jsdocs to reflect the order change
- Adds a `handleOhipKeyRollback` util function to perform an Ohip key deletion if needed. This is is only done if there is an error creating an Ohip number record in PHI DAS, and there is no existing record related to the Participant. Prisma doesn't provide a mechanism to determine if a record is newly created, so there is an extra `get` to PHI DAS here to determine that
- Adds a `deleteOhipKey` service call to support the rolback flow
- minor types + error return cleanup from demo work 

### Keys DAS
- adds `delete` endpoint and service call for Ohip router to support rollback flow
- fixes the `error.format()` call in one of the `GET /ohip-keys` logs

### PHI DAS
- adds an error clause for a conflict error in `updateOrCreateOhipNumberRecord`
- fixes some `error.format()` calls in the `ohip` and `clinical-profiles` router logs

### Types
- Added `hasRequiredOhipInfo` refinement to run a final check on the response object returned from `createConsentReleaseData` in Data Mapper

### Docs
- updates `create-consent-release-data-flow` diagram to reflect new order of operations
- adds a supplemental sequence diagram to describe the Ohip key rollback process

### Special Instructions
Pick up the new typing by running `pnpm run build` from the types package

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/keys-das/src/routers/ohipKeys.ts
```diff
@@ -179,4 +180,63 @@ router.post('/:participantId', async (req, res) => {
...
+ *       - OHIP Keys
+ *     name: Delete Ohip Key
+ *     description: Delete an Ohip Key record by Ohip Private Key
+ *     parameters:
+ *       - name: ohipPrivateKey
+ *         in: path
+ *         description: Ohip Private Key
+ *         required: true
+ *         schema:
+ *           type: string
```
mistryrn, line 197: Should we use the NanoId swagger schema here?

apps/keys-das/src/services/delete.ts
mistryrn, line 1: Awesome! 🎉

mistryrn (commented):
Great stuff!! 👏 I only had small comments about little documentation things, but other than that this looks awesome.


Let me know if you want to address the comments here or in the Cleanup ticket and then we can get this approved and merged 🙂

mistryrn (approved):
Looks good! Thanks for addressing the feedback so quickly! 🚀

Issue #707, opened 2024-06-13
Migrate Axios to fetch - consent UI -> API


samrichca:
For consent API requests in consent UI:
- Migrate Axios requests to fetch in order to use Next's cache & revalidation features
- Ensure data is being updated after mutations, when navigating between pages
  - Current bug: If a user's consent is in progress, and they navigate from the dashboard through the wizard and back to the dashboard in under 5 minutes, the dashboard will not update to show that consent is complete

PR #716, opened 2024-06-19
👽 💄 #707: axios on client side - consent step 3 & 4

samrichca:
## Description of Changes

We ran into issues making requests on the server side and keeping data in the UI fresh. In this PR I'm moving Axios requests to client side, and routing them through the proxy to keep server-side authentication, specifically for consent wizard step 3 & 4.

### Consent UI
- Remove server-side create requests for steps 3 & 4
  - Server-side get requests are still being used in step 5
- Handle null response bodies
- Add comment to qaLogger explaining that it's temporary for QA testing
- Add functions to get & create data through the Consent API proxy
- Add reusable hooks to manage get and post requests, and error & loading states
- Update consent steps 3 & 4 with the new requests
- Add form wrappers for step 3 & 4 to pass down pre-populated data.
- Add a type for GET functions, to use in form props
- Add VERBOSE_AXIOS_LOGGING to client app config
- Add NULL_BODY_RESPONSE_STATUSES to constants
- Add ConsentApiPath type, so we can specify that API request paths must be consent API paths
- Add useEffectOnce hook to stop useEffect hooks from firing multiple times, e.g. in dev/strict mode
- Add comments to server-side axios code.
- Add client-side axios interceptors and error handling.
- Add a combined consent API proxy fetch/client function, with axios interceptors.

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/hooks/useCreateConsentApiProxy.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: {
+	path: ConsentApiPath;
+	requestSchema: ZodSchema<V>;
+	responseSchema: ZodSchema<T>;
+}) => {
+	const [data, setData] = useState<T | undefined>(undefined);
+	const [error, setError] = useState<unknown | undefined>(undefined);
+	const [loading, setLoading] = useState<boolean>(false);
+
+	const createDataRequest = async (data: V): Promise<HandledResponse<T>> => {
```
joneubank, line 26: This function gets passed around quite a bit, for a specific example we can look at `ConsentRecontactFormWrapper` which passes it to `ConsentRecontactForm`. This means we need to use the **type** of this function quite a bit, like in the props definition for `ConsentRecontactForm`.

We should export a generic type to help us with this:

```ts
export type DataRequestHandler<RequestBody, ResponseBody> = (data: RequestBody) => Promise<HandledResponse<ResponseBody>>;
```

Then you can use this when declaring the props for a component:

```ts
const ExampleComponentWithDataSubmission= ({
	children,
	createHandler,
	createLoading,
}: {
	children: ReactNode;
	createHandler: DataRequestHandler<ExampleRequest, ExampleResponse>;
	createLoading: boolean;
}) => { 
  // ...
}
```

apps/consent-ui/src/constants/externalPaths.ts
```diff
@@ -38,7 +38,9 @@ export const API = {
 	PARTICIPANTS: '/participants',
 	STATUS: '/status',
 	USERS: '/users',
-};
+} as const;
+type ConsentApiPathKeys = keyof typeof API;
+export type ConsentApiPath = (typeof API)[ConsentApiPathKeys];
```
joneubank, line 43: We have a nice utility for this:

```ts
import type {  Values } from "types/common";

// ...

export type ConsentApiPath = Values<typeof API>;
```

apps/consent-ui/src/components/views/ConsentWizard/ConsentRecontact/ConsentRecontactForm/ConsentRecontactFormWrapper.tsx
```diff
@@ -0,0 +1,91 @@
...
+	formDict,
+	loadingText,
+}: {
+	children: ReactNode;
+	currentLang: ValidLanguage;
+	currentStep: ConsentStepRoute;
+	errorsDict: FormErrorsDictionary;
+	formDict: ConsentRecontactFormDictionary;
+	loadingText: string;
+}) => {
```
joneubank, line 57: Consider using `PropsWithChildren<{/* your props */}>` instead of manually declaring the type of the children.

apps/consent-ui/src/components/views/ConsentWizard/ConsentRecontact/ConsentRecontactForm/index.tsx
```diff
@@ -34,25 +39,30 @@ import { FormErrorsDictionary } from 'src/i18n/locales/en/formErrors';
...
 	formDict,
-	children,
 	currentStep,
 	loadingText,
 	data,
 }: {
+	children: ReactNode;
+	createHandler: (
+		data: ConsentRecontactFormRequest,
+	) => Promise<HandledResponse<CreateConsentRecontactResponse>>;
```
joneubank, line 61: See comment about creating a generic type to represent this handler function, will make reading the props much easier.

apps/consent-ui/src/components/views/ConsentWizard/ConsentResearchParticipation/ConsentResearchParticipationForm/index.tsx
```diff
@@ -72,28 +79,18 @@ const ConsentResearchParticipationForm = ({
...
-				);
-			}
-		});
+		const consentResearchParticipationResult = await createHandler(data);
+		if (consentResearchParticipationResult?.data) {
+			qaLogger(
+				'ConsentResearchParticipationForm',
+				'Created successfully',
+				consentResearchParticipationResult?.data || '',
+			);
```
joneubank, line 91: When I saw this I assumed it was a function that only created logs when we are in a non-production environment, presumably based on a config setting. The reality his that this is just very verbose logging that will also run in production. We should remove this logging or add controls that it will only occur in a dev environment.

apps/consent-ui/src/services/api/axios/client/catchAxiosError.ts
```diff
@@ -0,0 +1,60 @@
...
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+const DEFAULT_ERROR = 'Something went wrong';
```
joneubank, line 20: `DEFAULT_ERROR_MESSAGE`

apps/consent-ui/src/services/api/axios/client/catchAxiosError.ts
```diff
@@ -0,0 +1,60 @@
...
+			// `error.request` is an instance of XMLHttpRequest in the browser and an instance of
+			// http.ClientRequest in node.js
+			return String(error.request);
+		} else if ('message' in error) {
+			// Something happened in setting up the request that triggered an Error
+			return String(error.message);
+		} else {
+			return DEFAULT_ERROR;
+		}
+	}
```
joneubank, line 55: Pretty good unknown type handling here. Because we know our error is coming from axios we can simplify this quite a bit thanks to `axios.isAxiosError()`

```ts
import { isAxiosError } from 'axios';

// ...

	if (isAxiosError(error)) {
		if (error.response) {
			return `${error.response.status} - ${error.response.data}`;
		} else if (error.request) {
			return String(error.request);
		} else {
			return DEFAULT_ERROR;
		}
	}
```

apps/consent-ui/src/services/api/axios/client/catchAxiosError.ts
joneubank, line 1: There are two files in the codebase called `catchAxiosError`, can we combine to one?

apps/consent-ui/src/services/api/axios/client/axiosInterceptors.ts
joneubank, line 1: Can we add some comments to this file/functions to explain that these functions are only meant to be used in lower environments? I had originally commented that we shouldnt be logging like this in the browser but then i saw that they are only called when we have Verbose config enabled.

Since the control over when this logging is used is not included in the logging itself, we should add some comments to explain the purpose and how this is controlled... but also to warn future devs not to use them without similar controls.

joneubank: Alternately let these functions read from the config directly and only run their console.log if the verbose mode is enabled. Then you can always include these and they only do something based on the config.

apps/consent-ui/src/services/api/axios/client/consentApiProxyClient.ts
```diff
@@ -0,0 +1,61 @@
...
+import urlJoin from 'url-join';
+
+import {
+	axiosClientErrorInterceptor,
+	axiosClientRequestInterceptor,
+	axiosClientResponseInterceptor,
+} from './axiosInterceptors';
+
+const VERBOSE_AXIOS_LOGGING = process.env.NEXT_PUBLIC_VERBOSE_AXIOS_LOGGING === 'true';
+const CONSENT_UI_URL = process.env.NEXT_PUBLIC_CONSENT_UI_URL || '';
```
joneubank, line 32: Do we have config tools to use for this instead of reading directly from the `.env`? It is best to keep all of our environment variable references in a single place to simplify managing our configuration and environment. When the env interaction is spread throughout the app it becomes difficult to know how to configure it.

joneubank (changes requested):
This is amazing work and will be a big boon for future work. Consequently, I got a bit picky on some type points and naming so we can have a solid foundation to work off of. Looking forward to the final version (not a lot of changes to get there :) )

apps/consent-ui/src/hooks/useCreateConsentApiProxy.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { ConsentApiPath } from 'src/constants';
+import { qaLogger } from 'src/components/utils';
+import { createConsentApiProxy } from 'src/services/api/axios/client';
+
+/**
+ * React hook for making POST requests to Consent API from client components,
+ * through the authenticated proxy.
+ */
+const useCreateConsentApiProxy = <T, V>({
```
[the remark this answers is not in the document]
samrichca: good idea, we've been using `create` but I'd prefer `post` here because I'm not creating something, it's a generic post request.

apps/consent-ui/src/components/views/ConsentWizard/ConsentRecontact/ConsentRecontactForm/ConsentRecontactFormWrapper.tsx
samrichca: children is optional in PropsWithChildren so I explicitly define the children type when it's required

image

apps/consent-ui/src/components/views/ConsentWizard/ConsentResearchParticipation/ConsentResearchParticipationForm/index.tsx
anncatton: `qaLogger` was added just before our demo to help with qa testing, since we don't have full error handling yet. Not intended as a permanent feature! Might be good however to indicate this in function definition + a ticket `TODO`

apps/consent-ui/src/services/api/axios/client/catchAxiosError.ts
samrichca: one is for the server-side requests which I don't want to update while I'm transitioning to client-side requests

I made a ticket for revisiting catchAxiosError for any remaining server-side requests https://github.com/OHCRN/platform/issues/752

apps/consent-ui/src/services/api/axios/client/consentApiProxyClient.ts
samrichca: updated - getting the original env vars for consent UI URL & verbose logging from the hook, in the fetch hooks.

the NEXT_PUBLIC_ env vars have been removed

apps/consent-ui/src/services/api/axios/client/axiosInterceptors.ts
samrichca: I don't think I can have them read from the config directly because it's client-side code and the client config is provided with hooks.

I added comments explaining when & how to use these.

apps/consent-ui/src/services/api/axios/catchAxiosError.ts
```diff
@@ -17,6 +17,9 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+// TODO remove this when finished migrating server-side requests to client-side
+// https://github.com/OHCRN/platform/issues/752
```
mistryrn, line 21: Thanks for creating the ticket and linking to it here and in the comments

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -0,0 +1,170 @@
...
+	if (VERBOSE_AXIOS_LOGGING) {
+		client.interceptors.request.use(
+			(request) => axiosClientRequestInterceptor(request, AXIOS_CLIENT_NAME),
+			(error) => axiosClientErrorInterceptor(error, `${AXIOS_CLIENT_NAME} Request`),
+		);
+		client.interceptors.response.use(
+			(response) => axiosClientResponseInterceptor(response, AXIOS_CLIENT_NAME),
+			(error) => axiosClientErrorInterceptor(error, `${AXIOS_CLIENT_NAME} Response`),
+		);
+	}
```
mistryrn, line 83: Just thinking out loud here, but I wonder whether we need/want verbose axios logging for these requests, since it's just going from the client to our internal proxy 🤔 Open question: is there any value in keeping the verbose axios logging here?

Maybe we could remove this from the `consentApiProxyFetch` if it doesn't add much value? That way on line 86 below, you can just use `axios` similarly to how we're currently doing it in https://github.com/OHCRN/platform/blob/b70fbc05c8b1f95f2a6dd645f78e97c52a4e1ad8/apps/consent-ui/src/app/consent-api/%5B...proxy%5D/route.ts#L83

And speaking of the actual proxy handler itself, maybe that's something that _could_ potentially benefit from verbose logging, since those requests will be going to/from something external (Consent API)?

I'm also fine with leaving things as-is since this setup is currently working 😛 But if we feel some of this isn't needed/adding much value then we may as well simplify

mistryrn (commented):
Lots of great work in here 🙂 Only found one log that can be removed in the `useConsentApiProxyGet` hook – other than that it looks great!

I left a comment for a discussion on whether or not we want to keep the verbose axios logging for the requests going from the client to the proxy, let me know what you think!

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import catchAxiosError from './catchAxiosError';
+
+type ConsentApiProxyFetchArgs<TResponseBody> = {
+	baseUrl: string;
+	headers?: AxiosHeaders;
+	params?: string;
+	path: ConsentApiPath;
+	responseSchema: ZodSchema<TResponseBody>;
+	VERBOSE_AXIOS_LOGGING?: boolean;
```
mistryrn, line 37: Can this be removed now as well?

mistryrn (commented):
Looks good! Just one optional arg that can be removed from the `ConsentApiProxyFetchArgs` type I think 🙂

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
samrichca: thanks for catching that

mistryrn (dismissed):
Thanks for all the hard work that went into this one!

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -0,0 +1,150 @@
...
+	body,
+	headers,
+	method = 'GET',
+	params = '',
+	path,
+	responseSchema,
+}: ConsentApiProxyFetchArgs<TResponseBody> & {
+	body?: AxiosRequestConfig['data'];
+	method?: Method;
+}) => {
```
joneubank, line 60: It might be nice to declare the response type here:
```ts
export const consentApiProxyFetch = async <TResponseBody>({
	baseUrl,
	body,
	headers,
	method = 'GET',
	params = '',
	path,
	responseSchema,
}: ConsentApiProxyFetchArgs<TResponseBody> & {
	method?: Method;
}): Promise<Result<TResponseBody, 'SYSTEM_ERROR'>> => {
```

Reasons for this are:
- ensures there are no branches in this function that fail to return a Result
- ensures all successes return the correct data
- Makes it obvious what error conditions need to be handled.

joneubank (dismissed):
Looks great. One suggestion.

joneubank (approved):

PR #747, opened 2024-07-04
handle null body response statuses in route handler

anncatton:
## Description of Changes

Supplemental suggestions for https://github.com/OHCRN/platform/pull/716 . Specific handling for null body response http statuses

### Consent UI
- if response from consent-api request is any of `101, 103, 204, 205, 304` http status, the `Response` body is set to `undefined`. This fixes the 'Invalid Response status` error that occurred on `304`s
- adds `useEffectOnce` hook to ensure api hook requests do not occur multiple times on re-renders
- replacing use of `NEXT_PUBLIC` consent ui URL with the value held in `AppContext`. This means the baseUrl needs to be passed through the api hooks to be used by the underlying fetch function
- removes server action `GET` request from top level `.../ConsentWizard/ConsentResearchParticipation/index.tsx` file to prevent superfluous fetch

## PR Readiness Checklist

- [ ] "Expected Outcome(s)" in ticket have been met
- [ ] Ticket number included in PR title
- [ ] Connected ticket to PR
- [ ] Labels added to PR for service name (`consent-api`, `data-mapper`, etc...), type (`chore`, `documentation`, etc...), status (`draft`, `on-hold`, etc...) **if applicable**
- [ ] Manual testing completed
- [ ] Builds locally without errors or warnings


apps/consent-ui/src/app/consent-api/[...proxy]/route.ts
```diff
@@ -34,6 +34,7 @@ const regexMultipartFormData = new RegExp('multipart/form-data');
 const isFileUpload = (request: NextRequest) =>
 	regexMultipartFormData.test(request.headers.get('content-type') || '');
 
+const nullBodyResponseStatuses = [101, 103, 204, 205, 304];
```
Comment (mistryrn) [inline apps/consent-ui/src/app/consent-api/[...proxy]/route.ts:37]: For reference, this is from: https://fetch.spec.whatwg.org/#null-body-status – these are all the HTTP statuses which have null body responses.

We can probably make this a constant and rename it with `UPPER_SNAKE_CASE`, add a little comment above explaining/linking to the above resource for future reference. Probably wouldn't hurt to store it in the `constants/` directory, though I'm not sure at the moment if we'll use it elsewhere.

apps/consent-ui/src/components/views/ConsentWizard/ConsentResearchParticipation/index.tsx
```diff
@@ -42,11 +40,6 @@ const ConsentResearchParticipation = async ({ currentLang }: { currentLang: Vali
 
 	const studyConsentPdfUrl = urlJoin(ASSETS_PATH, CONSENT_PDFS_PATH, pageDict.studyConsentPdf);
 
-	const consentResearchParticipationResult = await handleGetConsentResearchParticipation();
```
mistryrn, line 45: For reference: this was where the mystery second consent-api request was coming from, even after switching to `useEffectOnce`

We'll need to remove these server-side requests from the other steps when we switch to client-side

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -49,11 +52,12 @@ const useConsentApiProxyGet = <TResponseBody,>({
 	const [data, setData] = useState<TResponseBody | undefined>(undefined);
 	const [error, setError] = useState<unknown | undefined>(undefined);
 	const [loading, setLoading] = useState<boolean>(true);
+	const { CONSENT_UI_URL } = useAppConfigContext();
```
mistryrn, line 55: In order to avoid adding those additional `NEXT_PUBLIC_` env vars to access the `CONSENT_UI_URL` from the client-side axios client (those env vars may not play nice with our deploy setup; the whole build vs runtime env var stuff from last fall), we're using the `appConfigContext` at the hook level.

The downside here is that we can no longer make a single axios client for these client-side axios requests (`consentApiProxyClient`) configured with the `baseUrl` set to `CONSENT_UI_URL`. Instead, we have to pass the `baseUrl` down as a prop.

As a consequence, we suggest getting rid of the `consentApiProxyClient` altogether, and having the function in `consentApiProxyFetch` just use axios since we're providing the `baseUrl` every time.

apps/consent-ui/src/hooks/useConsentApiProxyPost.tsx
```diff
@@ -65,6 +67,7 @@ const useConsentApiProxyPost = <TResponseBody, TRequestBody>({
 			path,
 			requestSchema,
 			responseSchema,
+			baseUrl: CONSENT_UI_URL,
```
mistryrn, line 70: As part of the clean-up here, we can put these in alphabetical order

apps/consent-ui/src/hooks/useEffectOnce.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef, useState } from 'react';
```
mistryrn, line 1: Need to add the copyright header still, plus we should add a comment with a link/reference to the source of the hook for future reference: https://blog.ag-grid.com/avoiding-react-18-double-mount/

apps/consent-ui/src/services/api/axios/client/consentApiProxyClient.ts
```diff
@@ -29,16 +28,11 @@ import {
...
 
 const AXIOS_CLIENT_NAME = 'consentApiProxyClient';
 
 const initAxiosClient = () => {
-	console.info(`Initializing ${AXIOS_CLIENT_NAME}...`);
-	return axios.create({
-		baseURL: urlJoin(CONSENT_UI_URL, 'consent-api'),
-		validateStatus: (status: number) => status >= 200 && status < 400,
-	});
+	return axios.create();
```
mistryrn, line 35: As mentioned in the comments on `useConsentApiProxyGet`, since we're no longer configuring a single client with the `baseUrl` here, this `consentApiProxyClient` is basically just the same as a regular axios instance now. We can safely remove this altogether and just use `axios` inside the functions in `consentApiProxyFetch`.

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -34,6 +34,7 @@ type ConsentApiProxyFetchArgs<TResponseBody> = {
 	params?: string;
 	path: ConsentApiPath;
 	responseSchema: ZodSchema<TResponseBody>;
+	baseUrl: string;
```
mistryrn, line 37: As part of cleanup, let's alphabetize these in each of the functions before merging.

apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -49,11 +50,12 @@ export const consentApiProxyFetch = async <TResponseBody>({
 	params = '',
 	path,
 	responseSchema,
+	baseUrl,
 }: ConsentApiProxyFetchArgs<TResponseBody> & {
 	body?: AxiosRequestConfig['data'];
 	method?: Method;
 }) => {
-	const reqUrl = urlJoin(path, params);
+	const reqUrl = urlJoin(baseUrl || '', 'consent-api', path, params);
```
mistryrn, line 58: This was a quick test just to see if we could get it working, but this can probably be reverted and up in the hooks (`useConsentApiProxyGet.tsx:60`, `useConsentApiProxyPost.tsx:70`) when we pass in `baseUrl`, we can set the base URL to `urlJoin(CONSENT_UI_URL, 'consent-api')`

https://github.com/OHCRN/platform/blob/39698fa1a954b857affa96f753385dcacbd87a89/apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx#L60

https://github.com/OHCRN/platform/blob/39698fa1a954b857affa96f753385dcacbd87a89/apps/consent-ui/src/hooks/useConsentApiProxyPost.tsx#L70

mistryrn (commented):
Adding some comments to give context to why we made some of these changes yesterday, plus some next steps for cleanup/refactoring 🙂

Issue #742, opened 2024-07-03
Registration Page: Migrate Data Fetching to Client Side (Consent UI --> API)


b-f-chan:
For consent API requests on the Registration page:

- [ ] Migrate Axios requests to work client-side and GET/POST to the proxy
- [ ] Ensure the proxy works with JSON requests
- [ ] Ensure data is being updated after mutations, when navigating between pages

samrichca:
Follow-up ticket for POST request: https://github.com/OHCRN/platform/issues/792

PR #763, opened 2024-07-17
👽 💄  #742 1/2: change register page GET /invites request to client-side

samrichca:
## Description of Changes

This PR updates the Register page to make client-side requests to the consent API proxy to request an invite if an `inviteId` was provided in the URL.

### Consent UI
- Change `RegisterFormWrapper` to be a client component & add API fetching
- Move translations to nearest server component
- Add `params` & `disabled` to `consentApiProxyFetch` for GET invites request
- Remove unused files for server-side requests

## Testing
- Go to `/en/invite` & make sure the browser console is open. 
- Create an invite.
- Get `inviteId` from console.
- Go to `/en/register?inviteId=MY_INVITE_ID`

## PR Readiness Checklist

- [x] Connected ticket to PR
- [ ] Builds locally without errors or warnings


apps/consent-ui/src/services/api/axios/client/consentApiProxyFetch.ts
```diff
@@ -60,7 +60,7 @@ export const consentApiProxyFetch = async <TResponseBody>({
 	body?: AxiosRequestConfig['data'];
 	method?: Method;
 }): ConsentApiProxyFetchResult<TResponseBody> => {
-	const reqUrl = urlJoin(baseUrl, path, params);
+	const reqUrl = urlJoin(baseUrl, `${path}${params}`);
```
samrichca, line 63: add params to URL

the path has to come from the consent API paths object, so params are separate

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -74,10 +81,18 @@ const useConsentApiProxyGet = <TResponseBody,>({
...
 
 		setLoading(false);
-	}, [CONSENT_UI_URL, path, responseSchema]);
+	}, [CONSENT_UI_URL, params, path, responseSchema]);
 
+	// get data if this hook is active.
+	// hooks cannot be called conditionally, so the condition is moved here.
 	useEffectOnce(() => {
-		getDataRequest();
+		if (isActive) {
```
samrichca, line 89: for our current purposes,  I think resetting this hook's state if the hook isn't meant to be used should be OK.

apps/consent-ui/src/components/views/Register/index.tsx
```diff
@@ -33,8 +31,29 @@ const Register = async ({
 	currentLang: ValidLanguage;
 	inviteId?: string;
 }) => {
-	const { translateNamespace } = getTranslation(currentLang);
+	const { translate, translateNamespace } = getTranslation(currentLang);
```
samrichca, line 34: RegisterFormWrapper was a server component. I made it a client component to match the consent wizard steps, and moved the translations up to Register/index.tsx

apps/consent-ui/src/components/views/Register/RegisterForm/RegisterFormWrapper.tsx
```diff
@@ -0,0 +1,124 @@
...
+	// format invite data for registration form,
+	// or reset invite data state if there's no invite.
+	useEffect(() => {
+		if (getData?.data) {
+			const formattedData = formatInviteDataForRegistration({
+				...getData.data,
+				// TODO inviteAccepted type isn't matching, even though I used the same type
+				// here and in formatInviteDataForRegistration.
+				// remove line 93 to see the TS error.
+				inviteAccepted: !!getData.data.inviteAccepted,
```
samrichca, line 93: I don't know what happened here with the types, if anyone has ideas let me know.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -44,9 +46,13 @@ import { useEffectOnce } from './useEffectOnce';
...
+	isActive = false,
+	params,
 	path,
 	responseSchema,
 }: {
+	isActive?: boolean;
+	params?: string;
 	path: ConsentApiPath;
 	responseSchema: ZodSchema<TResponseBody>;
 }) => {
```
joneubank, line 58: Should we add a similar `isActive` parameter to the POST version?

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -44,9 +46,13 @@ import { useEffectOnce } from './useEffectOnce';
  * 	return loading ? <LoadingPlaceholder /> : error ? <p>{error}</p> : <MyComponent data={data} />;
  */
 const useConsentApiProxyGet = <TResponseBody,>({
+	isActive = false,
```
joneubank, line 49: I think we should make this `true` by default. I think we would expect the hook to work by default, and this parameter is a mechanism to disable it. Also, we want the existing proxy get hooks to continue to send without having to go back through and set isActive to true on all of them.

joneubank (changes requested):
Updating review because i had a thought on the isActive implementation that should be reviewed.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
joneubank: I think I finally pinned down the issue I'm having with this interface... can we change this property to be `disabled` instead of `isActive`? Its a property we set to stop the hook from running, its not a check on whether the hook is active or not.

So invert the logic I described above, disabled would be `false` by default so that this hook runs unless it gets set to disabled.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
samrichca: definitely but I'm going to add it on a PR with a POST request

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
samrichca: good catch, default true was the intended behaviour

samrichca: on second thought, I'm going to change this hook so it returns a function we call as needed, like the post hook.

I want to use the get function conditionally here and based on user action on the dashboard page, so it'd be good to have more control

joneubank (approved):

Issue #743, opened 2024-07-03
Dashboard: Migrate Data Fetching to Client Side (Consent UI --> API)


b-f-chan:
For consent API requests on the Dashboard:

- [ ] Migrate Axios requests to work client-side and GET/POST to the proxy
- [ ] Ensure the proxy works with JSON requests
- [ ] Ensure data is being updated after mutations, when navigating between pages

samrichca:
- [ ] use the new /wizard/pdf endpoint
- [ ] remove other API calls related to the wizard

samrichca:
Might be able to merge this ticket with #606

PR #793, opened 2024-08-02
👽 💄 #743: update dashboard to use GET /wizard/pdf & client-side requests

samrichca:
## Description of Changes

This PR updates the Dashboard page to use client-side requests, and updates the consent PDF download button to use the new /wizard/pdf endpoint instead of making multiple API requests.

### Consent UI
- Rearrange client/server components in the Dashboard page to use client-side requests.
- Update Dashboard dictionary to contain all strings for that page
- Update GET hook so it can be used on demand (return data instead of only setting state, control whether loading is true on page load)
- Updated `createdAt: Date` to `signatureCreatedAt: string` to match API response
- Update mock data to match new structure
  - We're keeping mock data and testing buttons to facilitate PDF customization in the future
- Update API paths to add `/wizard/pdf`

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/components/views/Dashboard/ConsentPdfDownloadButton/index.tsx
```diff
@@ -17,50 +17,45 @@
...
-	const consentRecontactResult = await handleGetConsentRecontact();
-	const consentReviewSignResult = await handleGetConsentReviewSign();
 
-	const data: GenerateConsentPdfData = {
-		...mockDataParticipant,
-		...consentReleaseDataResult.data,
-		...consentResearchParticipationResult.data,
-		...consentRecontactResult.data,
-		...consentReviewSignResult.data,
+const ConsentPdfDownloadButton = ({
```
samrichca, line 32: I removed the wrapper since you can have a button and client-side request in the same component.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -44,18 +44,20 @@ import { useAppConfigContext } from 'src/components/providers/AppConfigContextPr
  * 	return loading ? <LoadingPlaceholder /> : error ? <p>{error}</p> : <MyComponent data={data} />;
  */
 const useConsentApiProxyGet = <TResponseSchema extends ZodSchema>({
+	loadingDefault = true,
```
joneubank, line 47: Wouldn't the default here be `false` since this function doesn't send the request until you call `getRequest`?  Having the ability to mark it as `loading=true` from first render (before the request starts) is nice to be able to do but this differs from how the hook works.. I am thinking of a future dev seeing loading=true and being confused why the request isn't sent.

If you keep it this way you may want to expand on the example code to show that they need to actually call `getRequest()`.

joneubank (commented):
Looks good. Left a comment to consider about the `loadingDefault` default value.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
samrichca: if I make `loading=false` by default I get a flash of un-loaded content before the loading spinner comes up. I updated the examples and comments though, let me know if it's clearer now.

joneubank (approved):

PR #795, opened 2024-08-07
👽 💄 #738: update Consent Wizard step 2 to use client-side requests

samrichca:
## Description of Changes

This PR updates Consent Wizard step 2 to use client-side requests, removes unused code for server-side requests, and renames some files and variables for i18n to match other i18n uses.

### Consent UI
- Add client-side requests to step 2
- Remove server-side request files
- Add hidden field for `hasOhip` for backend, that always has the opposite value of `ohipDisabled`

### Types
- Add `transform()` to optional fields to transform empty strings to undefined, similar to `participantPreferredName`
- Add `hasOhip` to UI form type

## PR Readiness Checklist

- [x] Connected ticket to PR


packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
```diff
@@ -97,6 +107,7 @@ export type OhipFormInfo = z.infer<typeof OhipFormInfo>;
 
 export const ConsentReleaseDataFormRequest = ConsentReleaseDataFormBase.and(SelfIdentifiedGender)
 	.and(SelfReportedPrimaryCancerDiagnosis)
-	.and(OhipFormInfo);
+	.and(OhipFormInfo)
+	.and(z.object({ hasOhip: z.boolean() }));
```
samrichca, line 111: trying to remove `ohipDisabled` from the request using `transform(... undefined)` didn't work because that field has a `refine` function. however sending 1 extra field to the API doesn't break the request. I believe this is the only case where we have UI fields that don't exist in the backend.

packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
```diff
@@ -52,11 +52,21 @@ export const ConsentReleaseDataFormBase = ParticipantBaseOhipNameFields.merge(
 			ancestry: Ancestry,
 			familyHistoryOfCancer: HistoryOfCancer,
 			residentialPostalCode: PostalCode,
-			participantOhipMiddleName: EmptyOrOptionalName,
-			participantPreferredName: EmptyOrOptionalName,
-			selfReportedClinicianTitleOrRole: EmptyOrOptionalString,
-			selfReportedClinicianFirstName: EmptyOrOptionalName,
-			selfReportedClinicianLastName: EmptyOrOptionalName,
+			participantOhipMiddleName: EmptyOrOptionalName.transform(
```
samrichca, line 55: added these `transform` functions to stop getting 400 errors back from the API

packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
```diff
@@ -52,11 +52,21 @@ export const ConsentReleaseDataFormBase = ParticipantBaseOhipNameFields.merge(
...
+				(input) => input?.trim() || undefined,
```
joneubank, line 56: Maybe you want to define this function and reuse the named variable?

```ts
const transformFormString = (input: string) => input?.trim() || undefined;
```

so that we can just do 

```ts
participantPreferredName: EmptyOrOptionalName.transform(transformFormString),
```

apps/consent-ui/src/components/views/ConsentWizard/ConsentReleaseData/ConsentReleaseDataForm/ConsentReleaseDataFormWrapper.tsx
```diff
@@ -0,0 +1,130 @@
...
+	errorsDict: FormErrorsDictionary;
+	formDict: ConsentReleaseDataFormDictionary;
+	genderOptions: GenderOption[];
+	geneticsClinicOptions: GeneticsClinicOption[];
+	historyOfCancerOptions: HistoryOfCancerOption[];
+	loadingText: string;
+	molecularLabOptions: MolecularLabOption[];
+	primaryCancerDiagnosisOptions: PrimaryCancerDiagnosisOption[];
+	textDict: ConsentReleaseDataTextDictionary;
+}) => {
```
joneubank, line 84: I don't know what to suggest immediately but... this is a lot of props... clearly this isn't a reusable component and its tightly coupled to app state. We should think a bit about how to manage this much state and if passing it through props is what we want to do.

joneubank (approved):

packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
mistryrn: I think this approach is fine, it seems to be working well. I guess the new client-side request setup doesn't have room for the types of transformation functions we were applying to the data before submitting requests in our server-side setup? Like `transformConsentReleaseData` https://github.com/OHCRN/platform/pull/795/files?diff=unified&w=0#diff-0d63b359572e330eff9c0952d63178f1457cda6c1ad888c4f402552259621a2eL32

packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
mistryrn: I like this suggestion! 🙂

mistryrn (commented):
Looks good to me! 🙂 🎉  Tested locally and working as expected. I liked Jon's suggested refactoring for that transform, but not going to block merging for this PR for that.

apps/consent-ui/src/components/views/ConsentWizard/ConsentReleaseData/ConsentReleaseDataForm/ConsentReleaseDataFormWrapper.tsx
samrichca: it's all i18n dictionaries, if that helps.

packages/types/src/services/consentUi/requests/CreateConsentReleaseData.ts
samrichca: at the moment the request and the form have to have the same schema, which is OK in all of the forms except this one because the backend equivalent for this one field has to have the opposite value of the UI field. I think we should change this though in case it's needed in the future. I'm going to take a stab at it, I'll loop jon into a zoom call if I can't figure it out.

Issue #799, opened 2024-08-08
Add allowlist for institutional emails


samrichca:
- [ ] Add allowlist for institutional emails e.g. in an enum
- [ ] Apply allowlist to clinician institutional email fields in Invite and Step 2
- [ ] Show error if the email address doesn't fit the allowlist

PR #810, opened 2024-08-20
🏷️ #799: add institutional email allowlist

samrichca:
## Description of Changes

This PR adds an allowlist for institutional email address domain names, so that clinician invites can only be initiated by people with appropriate email addresses.

### General
- Update vitest to continue using the vscode plugin

### Consent API
- Add refinement to `POST /invites` schema to check email address domains

### Consent UI
- Add refinement to schemas for the invite form & post request to check email address domains

## Testing
Make sure to update env file in consent API & UI.

## PR Readiness Checklist

- [x] Connected ticket to PR


packages/express-request-validation/tests/index.spec.ts
```diff
@@ -56,7 +56,7 @@ describe('withRequestValidation', () => {
 	});
 
 	it('Valid request - should run request handler with body content that still passes validation', () => {
-		const requestHandler = vi.fn<Parameters<RequestHandler>, ReturnType<RequestHandler>>((req) => {
```
samrichca, line 59: this produced an error when I upgraded to vitest 2

package.json
```diff
@@ -47,11 +47,11 @@
 		"@types/node": "^20.1.3",
 		"@types/swagger-jsdoc": "^6.0.1",
 		"@types/swagger-ui-express": "^4.1.4",
-		"@vitest/coverage-istanbul": "^1.6.0",
+		"@vitest/coverage-istanbul": "^2.0.5",
 		"prettier": "^2.8.8",
 		"ts-node": "^10.9.1",
 		"typescript": "^5.4.5",
-		"vitest": "^1.6.0"
+		"vitest": "^2.0.5"
```
samrichca, line 54: this update is needed to continue using the vitest vscode plugin.

packages/types/src/common/utils/clinicianInstitutionalEmail.ts
```diff
@@ -0,0 +1,59 @@
...
+	'shn.ca',
+	'sickkids.ca',
+	'sinaihealth.ca',
+	'sunnybrook.ca',
+	'tbh.net',
+	'trilliumhealthpartners.ca',
+	'uhn.ca',
+	'wchospital.ca',
+	'wrh.on.ca',
+];
```
joneubank, line 45: This is hardcoded into a service that is shared across every module application we run. If we need to change this list, we're going to need to rebuild and deploy all of our services that validate emails.

To some extent this is unavoidable... but I think it may be worth moving this validation only onto the API service. We only need to validate emails as they are submitted to us, once they are inside the system we should accept them without rechecking the domain.

Would you consider changing the approach her to just being an input validation that is done on specific api endpoints instead of modifying our email rules?

joneubank (changes requested):

apps/consent-api/src/routers/invites.ts
```diff
@@ -43,7 +44,14 @@ import { getInvite } from '../services/search.js';
 
 const router = Router();
 
-const ClinicianInviteSchema = z.object({ data: ClinicianInviteRequest });
+const ClinicianInviteSchema = z.object({
+	data: ClinicianInviteRequest.refine(checkClinicianInstitutionalEmailDomain, {
+		// this refinement is added here in order to get the email domain allowlist
+		// from environment variables at runtime.
+		message: 'Clinician institutional email domain is not allowed',
```
anncatton, line 51: we can still put the error in the `refinementErrorMessages.ts` file

anncatton (commented):
Added some comments + the prisma version fix merged to develop created a conflict in the lockfile.

apps/consent-api/src/routers/invites.ts
```diff
@@ -48,7 +51,7 @@ const ClinicianInviteSchema = z.object({
 	data: ClinicianInviteRequest.refine(checkClinicianInstitutionalEmailDomain, {
 		// this refinement is added here in order to get the email domain allowlist
 		// from environment variables at runtime.
-		message: 'Clinician institutional email domain is not allowed',
+		message: clinicianInstitutionalEmailDomainNotAllowed,
```
anncatton, line 54: 🙏

anncatton (approved):
thanks for the updates @samrichca !

apps/consent-api/.env.schema
```diff
@@ -18,3 +18,6 @@ FILE_STORAGE_REGION=us-east-1
 
 # Developers can use the Recaptcha V2 test key found here: https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
 RECAPTCHA_SECRET_KEY=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+
+# Space-separated list of allowed domains for clinician institutional email addresses
+CLINICIAN_INSTITUTIONAL_EMAIL_DOMAIN_ALLOWLIST=
```
joneubank, line 23: You may want to include an example value to show how to format this list.

apps/consent-api/src/utils/checkClinicianInstitutionalEmailDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @returns boolean
+ */
+export const checkClinicianInstitutionalEmailDomain = ({
+	clinicianInstitutionalEmailAddress,
+}: {
+	clinicianInstitutionalEmailAddress: Email;
+}) => {
+	const { clinicianInstitutionalEmailDomainAllowList } = getAppConfig();
+	const allowList = clinicianInstitutionalEmailDomainAllowList.split(' ');
+	return allowList.includes(clinicianInstitutionalEmailAddress.split('@')[1] || '');
```
joneubank, line 38: If the env value is formatted poorly, for example two spaces back to back like `gmail.com  yahoo.com`, then the domain `''` will be allowed.

This isn't particularly bad but its an incomplete check. We should better validate and check our env value by filtering out empty values. We should also perform this parsing on start up, theres no reason to perform the split operation every time a user submits a form.

joneubank (dismissed):
Works fine. Left suggestion for improvement.

apps/consent-api/.env.schema
mistryrn: I like Jon's suggestion of including example values, maybe we can include some example/dummy domains in here to make local setup and testing easier?

Looks like `example.com`, `example.net`, `example.org` etc are reserved domain names that we can use without worry: https://en.wikipedia.org/wiki/Special-use_domain_name#Reserved_domain_names

apps/consent-ui/src/services/api/axios/client/catchAxiosError.ts
```diff
@@ -34,7 +34,7 @@ const catchAxiosError = (error: unknown): string => {
 
 	if (isAxiosError(error)) {
 		if (error.response) {
-			return `${error.response.status} - ${error.response.data}`;
+			return `${error.response.status} - ${JSON.stringify(error.response.data)}`;
```
samrichca, line 37: fix issue where errors were returning e.g. `401 - [Object object]`

mistryrn (commented):
Looks good! I agree with Jon's comments, left a suggestion for the examples.

mistryrn (approved):
Thanks for updating, looks good!

PR #819, opened 2024-08-29
💄 👽 #823: refactor consent API proxy hooks to use a reducer

samrichca:
## Description of Changes

This PR updates the consent API proxy hooks to use a reducer to make the functionality more stable compared to using multiple useState functions.

### Consent UI
- Refactor useConsentApiProxyGet & useConsentApiProxyPost to use a reducer

## PR Readiness Checklist

- [x] Connected ticket to PR


apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -26,6 +26,20 @@ import { ConsentApiPath, PROXY_API_PATH } from 'src/constants';
...
+	loading: boolean;
+};
+
+type ReducerAction<TResponseBody> =
+	| {
+			type: 'reset';
+	  }
+	| { type: 'init'; payload: { loading: boolean } }
+	| { type: 'success'; payload: { data: TResponseBody } }
+	| { type: 'error'; payload: { error: any } };
```
samrichca, line 39: if I change `error: any` to `error: unknown`, I get errors

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -60,10 +74,37 @@ const useConsentApiProxyGet = <TResponseSchema extends ZodSchema>({
 	path: ConsentApiPath;
 	responseSchema: TResponseSchema;
 }) => {
-	type ReponseBody = Zod.infer<TResponseSchema>;
-	const [data, setData] = useState<ReponseBody | undefined>(undefined);
-	const [error, setError] = useState<unknown | undefined>(undefined);
-	const [loading, setLoading] = useState<boolean>(loadingDefault);
+	type ResponseBody = Zod.infer<TResponseSchema>;
+
+	const reducer = (
```
samrichca, line 79: I tried to move this outside of the hook, but I had trouble getting the generics to work.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -75,8 +116,8 @@ const useConsentApiProxyGet = <TResponseSchema extends ZodSchema>({
 	 * To get data on page load: use this function in a `useEffect` or `useEffectOnce` hook.
 	 * To get data after page load, ex. after click a button: Set `loadingDefault` to `false` and use `getRequest` in the button click handler function.
 	 */
-	const getRequest = useCallback(async (): Promise<ReponseBody | undefined> => {
-		setLoading(true);
+	const getRequest = useCallback(async (): Promise<ResponseBody | undefined> => {
```
samrichca, line 119: typo

packages/express-request-validation/tests/index.spec.ts
```diff
@@ -56,7 +56,7 @@ describe('withRequestValidation', () => {
 	});
 
 	it('Valid request - should run request handler with body content that still passes validation', () => {
-		const requestHandler = vi.fn<Parameters<RequestHandler>, ReturnType<RequestHandler>>((req) => {
+		const requestHandler = vi.fn((req) => {
```
samrichca, line 59: fixing build issue with vitest upgrade

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
```diff
@@ -26,6 +26,20 @@ import { ConsentApiPath, PROXY_API_PATH } from 'src/constants';
 import { consentApiProxyGet } from 'src/services/api/axios/client';
 import { useAppConfigContext } from 'src/components/providers/AppConfigContextProvider';
 
+type ReducerState<TResponseBody> = ({ data: TResponseBody } | { error: unknown }) & {
+	loading: boolean;
+};
```
joneubank, line 31: What is the value of this state before any data is loaded? It would have no data or error, which is only legal according to this definition since `error: unknown` is satisfied by `undefined`.

I might suggest instead:
```
type ReducerState<TResponseBody> =
	| {
			loading: true;
	  }
	| ({
			loading: false;
	  } & ({ data: TResponseBody } | { error: unknown }) & {
				loading: boolean;
			});
```

And set the default to:

```ts
const reducerDefaultState = { loading: true };
```

This works so long as you don't need a way to differentiate "request not started" and "request currently underway". If you do, then consider using state as an enum with options `'LOADING' | 'ERROR' | 'SUCCESS' | 'NOT_STARTED'` and a type like:

```ts
type RequestState = 'LOADING' | 'ERROR' | 'SUCCESS' | 'NOT_STARTED';
type ReducerState<TResponseBody> =
	| {
			state: 'NOT_STARTED'
	  }
	| {
			state: 'LOADING'
	  }
	| {
			state: 'ERROR',
			error: unknown
	  }
	| {
			state: 'SUCCESS',
			data: TResponseBody
	  };
```

joneubank (changes requested):
This is a significant improvement, I like this model a lot.

apps/consent-ui/src/hooks/useConsentApiProxyGet.tsx
samrichca: I like the idea of states. 

BTW most of our get requests happen right away on page load and some are triggered by the user, like consent PDF, so we have to account for both

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { z, ZodSchema } from 'zod';
+import urlJoin from 'url-join';
+
+import { qaLogger } from 'src/components/utils';
+import { ConsentApiPath, PROXY_API_PATH } from 'src/constants';
+import { consentApiProxyGet } from 'src/services/api/axios/client';
+import { useAppConfigContext } from 'src/components/providers/AppConfigContextProvider';
+
+const REDUCER_STATUSES = ['NOT_STARTED', 'LOADING', 'SUCCESS', 'FAILURE'] as const;
+export const ReducerStatus = z.enum(REDUCER_STATUSES);
```
samrichca, line 30: I changed this to `status` to avoid conflicts with `state`

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
```diff
@@ -0,0 +1,174 @@
...
+export type ReducerStatus = z.infer<typeof ReducerStatus>;
+
+type ReducerState<TResponseBody> =
```
samrichca, line 33: `data` and `error` have to be specified for all versions of state, or else when I use this hook I get errors in vscode that data & error aren't known properties.

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
```diff
@@ -0,0 +1,174 @@
...
+	| ({ data: undefined; error: undefined } & ({ status: 'NOT_STARTED' } | { status: 'LOADING' }))
+	| { status: 'SUCCESS'; data: TResponseBody; error: undefined }
+	| { status: 'FAILURE'; data: undefined; error: any };
```
joneubank, line 36: If a property is `undefined`, just don't include the property (`error` and `data` respectively).

joneubank: Is this done to simplify the reducer assignment/overwrite? Does the reducer only replace the values specified instead of completely replacing the stored value with the new object?

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
```diff
@@ -0,0 +1,174 @@
...
+					error: action.payload,
+					status: ReducerStatus.enum.FAILURE,
+				};
+				break;
+			case 'SUCCESS':
+				result = {
+					data: action.payload,
+					error: undefined,
+					status: ReducerStatus.enum.SUCCESS,
+				};
```
joneubank, line 108: Maybe this is a limitation of the reducer interface that I am about to learn (ie. maybe its not an overwrite but a merge value)... couldn't we just define the result as:

```ts
result = {
					data: action.payload,
					status: ReducerStatus.enum.SUCCESS,
				};
```

joneubank (approved):
I think this is fine... The reducer syntax cause us to declare the properties with type `undefined` is really awkward but not strictly wrong.

anncatton (commented):
Are these changes connected to a ticket?

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
samrichca: when I left out undefined properties I got errors saying the properties were unknown.

the reducer overwrites all values

samrichca:
> Are these changes connected to a ticket?

it is now ;)

apps/consent-ui/src/hooks/useConsentApiProxyGetReducer.tsx
samrichca: if I don't make data & error known properties at all times, I get type errors:

image

joneubank (approved):

PR #829, opened 2024-09-06
Add `nx` monorepo manager

joneubank:
## Description of Changes

Add [nx](https://nx.dev/) as a monorepo manager. This has been added to support changes to the Jenkins pipeline as nx provides tools to determine which apps are affected by changes in the commit, including checking dependency trees for apps affected by changes in their dependent packages.

This has some bonus improvements for local developer work experience by caching builds and ensuring all dependencies get built when required.


Jenkinsfile
```diff
@@ -1,2 +1,2 @@
-@Library(value='jenkins-pipeline-library@master', changelog=false) _
-pipelineOHCRNPlatform()
\ No newline at end of file
+@Library(value='jenkins-pipeline-library@feat/ohcrn-monorepo-separate-project-stages', changelog=false) _
```
joneubank, line 1: change back to master once testing is done, before merging.

apps/consent-api/src/index.ts
```diff
@@ -72,3 +72,5 @@ const App = (config: AppConfig) => {
 };
 
 export default App;
+
+// Comment to trigger change
```
mistryrn, line 76: Do we still need this, or should we remove it in a subsequent PR after the merge?

mistryrn (approved):
Thanks for tackling this! Looks great, just one comment about the comment left in the consent-api `index.ts`

apps/consent-api/src/index.ts
joneubank: oops, thanks for noticing.

mistryrn (approved):
🚀

Issue #736, opened 2024-07-02
Consent Wizard Step 2: "I have no OHIP" Should Not Be Selected by Default


b-f-chan:
There is bug on Consent Wizard Step 2, where the "I have no OHIP number" checkbox is selected by default on page entry/load

This is WRONG --> It should be unselected by default, as most users will have an OHIP number

PR #872, opened 2024-09-18
#736 OHIP checkbox not checked by default now

shailikadakia-oicr:
## Description of Changes
Currently, when a user is on Step 2 of the consent wizard (2. Consent to Data Release page), the default behaviour is assuming users don't have an OHIP number. As a result, the checkbox with label `I do not have an Ontario health card` is checked off.  We don't want this to be the default behaviour as most users will have an OHIP card. 

Updated the default value of `ohipDisabled` to `false`. `ohipDisabled` and `hasOhip` are opposite to each other so `hasOhip` defaults to true for the rest of the form. 

#### `ConsentReleaseDataForm`
- When the react-hook-form is set up, the default values is updated to`ohipDisabled: false`. Originally, it was taking the opposite value of `hasOhip` through optionalChaining. We can keep the existing format and update to `ohipDisabled: !data?.hasOhip === false`. 
@mistryrn what do you think? Would this cause side effects? 

## PR Readiness Checklist

- [x] Connected ticket to PR
- [ ] [Axe DevTools](https://www.deque.com/axe/browser-extensions/) "Full Page Scan" of updated pages passes without introducing new Critical or Serious issues (UI changes only)


mistryrn (commented):
This works fine for the user's first pass through the form, but I'd be concerned about side effects for if a user actually doesn't have an OHIP, selects the checkbox, proceeds to the next steps and returns to Step 2 to edit something.

In that case, would this ignore what the user had selected?

~Maybe we can add a conditional here that checks to see whether the user has submitted Step 2 previously? For now we can leverage one of the other required fields from Step 2 that will exist if the user returns to the form, like "Personal History of Cancer", something like:~
~ohipDisabled: data?.historyOfCancer ? !data?.hasOhip : false,~

Edit: we should just be able to use `data.ohipNumber` 😅 
```tsx
ohipDisabled: !data?.ohipNumber,
```

Try that out and let me know how it works, ideally we want:
- first time through Step 2: false
- returning to Step 2 after submitting it: opposite of `hasOhip`

apps/consent-ui/src/components/views/ConsentWizard/ConsentReleaseData/ConsentReleaseDataForm/index.tsx
```diff
@@ -105,7 +105,7 @@ const ConsentReleaseDataForm = ({
 		mode: 'onBlur',
 		defaultValues: {
 			...data,
-			ohipDisabled: !data?.hasOhip,
+			ohipDisabled: false,
```
samrichca, line 108: I think checking for user data first before deciding on a value would be a good direction here

mistryrn:
> ohipDisabled: data?.hasOhip !== undefined ? !data?.hasOhip : false,

(from https://github.com/OHCRN/platform/pull/872#issuecomment-2361696384)

Great suggestion, @samrichca ! @shailikadakia-oicr let's go with this after you get the chance to test it 🙂

mistryrn (approved):

